PR #28, opened 2017-09-20
insert columns in sorted order

lehnerchristian:
Add columns to a ranking in a sorted order. If there are already columns with the selected ID in the ranking, the new column will be inserted between the last occurrence of the selected ID and before the next ID
If the selected ID was not found the new column is inserted at the end of the ranking

### DIFF
--- a/src/lineup/ARankingView.ts
+++ b/src/lineup/ARankingView.ts
@@ -279,7 +279,7 @@ export abstract class ARankingView extends AView {
     this.fire(AView.EVENT_UPDATE_ENTRY_POINT, namedSet);
   }
 
-  private addColumn(colDesc: any, data: Promise<IScoreRow<any>[]>, id = -1, position): { col: Column, loaded: Promise<Column> } {
+  private addColumn(colDesc: any, data: Promise<IScoreRow<any>[]>, id = -1, position?: number): { col: Column, loaded: Promise<Column> } {
     const ranking = this.provider.getLastRanking();
 
     //mark as lazy loaded
--- a/src/lineup/ARankingView.ts
+++ b/src/lineup/ARankingView.ts
@@ -292,13 +292,12 @@ export abstract class ARankingView extends AView {
 
     this.provider.pushDesc(colDesc);
 
-    let col: Column;
+    const col: Column = this.provider.create(colDesc);
     if(lastIndex === -1) {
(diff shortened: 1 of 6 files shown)


src/lineup/ARankingView.ts
```diff
@@ -287,8 +287,21 @@ export abstract class ARankingView extends AView {
     colDesc.color = this.colors.getColumnColor(id);
     const accessor = createAccessor(colDesc);
 
+    // find last index of the current selected ID
```
sgratzl: have you considered adding some logic to the (multi) selection adapter itself? in the end this is the one that cares about the order

src/lineup/ARankingView.ts
```diff
@@ -287,8 +287,21 @@ export abstract class ARankingView extends AView {
...
+    const lastIndex = ranking.flatColumns.map((col) => (<any>col.desc).selectedId).lastIndexOf(id);
```
sgratzl: `.flatColumns` is a quite expensive operation, you could search for the previous column directly instead of the index

src/lineup/ARankingView.ts
```diff
@@ -287,8 +287,21 @@ export abstract class ARankingView extends AView {
...
+
     this.provider.pushDesc(colDesc);
-    const col = this.provider.push(ranking, colDesc);
+
+    let col: Column;
+    if(lastIndex === -1) {
+      // insert the column at the end of the ranking
+      col = this.provider.push(ranking, colDesc);
```
sgratzl: alternative version like the create below and then `ranking.push(col)` Then you don't have two different logic how to add a column

src/lineup/ARankingView.ts
lehnerchristian: yes, I've considered that, but I didn't see a better place, since I need to find the previous column after which I want to insert my current column. As far as I have seen it I only have column descriptions before
Can you point me to a place which would be better to sort the columns beforehand?

src/lineup/ARankingView.ts
lehnerchristian: you mean using the `ranking.find(string | (Column) => boolean)` method? because it uses the flatColumns method too
 I thought of calling the `flatColumns` method only once and saving the result to a variable
what would be another method I could use?

PR #37, opened 2017-10-13
Remember filter settings

lehnerchristian:
closes Caleydo/tdp_bi_bioinfodb#692

### DIFF
--- a/src/form/internal/AFormElement.ts
+++ b/src/form/internal/AFormElement.ts
@@ -72,7 +72,7 @@ export abstract class AFormElement<T extends IFormElementDesc> extends EventHand
     this.$node.classed('hidden', !visible);
   }
 
-  protected build() {
+  protected addChangeListener() {
     if (this.desc.useSession || this.desc.onChange) {
       this.on(AFormElement.EVENT_CHANGE, () => {
         this.updateStoredValue();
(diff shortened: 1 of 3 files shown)


lehnerchristian:
I found an issue with that. the session value needs to be reset when clicking on the `x` (close) button in a row. I'll fix it and comment again, when it's ready to review

sgratzl:
isn't the main issue that the following method won't be called for FormMaps?

https://github.com/datavisyn/tdp_core/blob/0e43af866398906d4e89b4e8bd73304776fdb606/src/form/internal/AFormElement.ts#L75-L86

lehnerchristian:
right, that's the issue

sgratzl:
there was a reason why the super method wasn't called. Does it still work for both the inline and regular (=score) case?

lehnerchristian:
yes, for me it's working. the filter settings are remembered, as the issue says, which is addressed by this issue. or was the reason that aggregated score dialogs should not remember the filters?
test

sgratzl:
> yes, for me it's working. the filter settings are remembered, as the issue says, which is addressed by this issue. or was the reason that aggregated score dialogs should not remember the filters?

take a close look -> the filter label is twice since both the `super.build` and `build` adds it.

PR #47, opened 2017-11-09
Lineup dev

sgratzl:
with the new release and the pretty stable lineup dev we can think of merging it


src/internal/PhoveaDataAdapter.ts
```diff
@@ -0,0 +1,56 @@
...
+      const vi = <INumberValueTypeDesc>v;
+      return {
+        type: <'number'>'number',
+        min: vi.range[0],
+        max: vi.range[1]
+      };
+    case 'categorical':
+      const vc = <ICategoricalValueTypeDesc>v;
+      return {
+        type: <'categorical'>'categorical',
```
lehnerchristian: is this some form of TS feature? a type cast?

sgratzl: yes, some kind of cast. you know you can write e.g. pseudo enum, like `test(v: 'a'|'b')`. this is the corresponding cast to convert to the type `'a'`. Needed it since it would be by default just a string.

PR #63, opened 2017-12-05
add possibility to not call the selection adapter and add a column (e…

lehnerchristian:
….g. if lineup is removed and rebuilt completely)


sgratzl:
I'm not sure whether it is so easy to change the interface of `parameterChanged`, since all overridden methods would take care of the extra parameter when calling the super method

PR #65, opened 2017-12-11
improved lineup color management

lehnerchristian:
- count how many items have the color before clearing it from the map
- avoid removing the color and then adding it back to the available colors, instead get the diff of the arrays and then get the first color

### DIFF
--- a/src/lineup/internal/LineUpColors.ts
+++ b/src/lineup/internal/LineUpColors.ts
@@ -1,11 +1,12 @@
 import Ranking from 'lineupjs/src/model/Ranking';
 import {scale} from 'd3';
+import {array_diff} from './LineUpSelectionHelper';
 
 export default class LineUpColors {
   /**
    * Map that assigns each selection ID a color, which is used as color for columns
    */
-  private readonly colorMap = new Map<number, string>();
+  private readonly colorMap = new Map<number, {color: string, items: number}>();
   private colors: string[] = scale.category10().range().slice();
 
   init(ranking: Ranking) {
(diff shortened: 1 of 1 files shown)


src/lineup/internal/LineUpColors.ts
```diff
@@ -24,22 +25,30 @@ export default class LineUpColors {
...
 
   freeColumnColor(id: number): void {
-    const color = this.colorMap.get(id);
-    if (color) {
-      this.colorMap.delete(id);
-      if (this.colors.indexOf(color) === -1) {
-        this.colors.push(color);
+    const value = this.colorMap.get(id);
+
+    if (value.color) {
```
sgratzl: will result in an error if `id` is not in the colorMap

src/lineup/internal/LineUpColors.ts
```diff
@@ -24,22 +25,30 @@ export default class LineUpColors {
...
+      value.items--;
+      if (value.items === 0) {
+        this.colorMap.delete(id);
+      } else {
+        this.colorMap.set(id, value);
```
sgratzl: since by reference setting it again is not needed

src/lineup/internal/LineUpColors.ts
```diff
@@ -1,11 +1,12 @@
...
 import {scale} from 'd3';
+import {array_diff} from './LineUpSelectionHelper';
 
 export default class LineUpColors {
   /**
    * Map that assigns each selection ID a color, which is used as color for columns
    */
-  private readonly colorMap = new Map<number, string>();
+  private readonly colorMap = new Map<number, {color: string, items: number}>();
   private colors: string[] = scale.category10().range().slice();
```
sgratzl: please extent as in https://github.com/Caleydo/lineupjs/blob/develop/src/model/CategoricalColumn.ts#L18

PR #83, opened 2018-02-14
Notifications

### LINKED ISSUE #81: Trigger notifications or alerts on (un)successful events
As a user, I want to get a feedback whether an action in the user interface has been (un)successful.

For example simple bootstrap alerts: see https://v4-alpha.getbootstrap.com/components/alerts/#dismissing

sgratzl:
fix #81 

provides more user feedback in successful and erroneous situations by using notifications (success will disappear automatically) 

2018-02-14 12_38_51-dtiles

replaces the "showModialDialogOnError" with an danger alert message


thinkh:
Good idea!

I would keep the success messages a second longer, because it was hard to locate and read the message for the first few times.

When I delete multiple (temporary) sessions fast enough (within in the 3 seconds) then they will not be deleted. Probably the variable for the timer is overridden.

ordino-notification-close-timer

I couldn't provoke an error now. 😄 However, I assume that they will stacked like the success messages. Correct?

src/notifications.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }, 10); // wait dom rendered
+}
+
+export function successfullySaved(type: string, name: string) {
+  pushNotification('success', `${type} "${name}" successfully saved`, 5000);
+}
+
+export function successfullyDeleted(type: string, name: string) {
+  pushNotification('success', `${type} "${name}" successfully deleted`, 3000);
```
thinkh: I would set the timer for the success message to 4000 or maybe to 5000.

sgratzl:
we can merge this PR after the release

PR #91, opened 2018-03-08
Sluger/select3

sluger:
[x] compiles with tslint

### DIFF
--- a/src/form/internal/FormSelect3.ts
+++ b/src/form/internal/FormSelect3.ts
@@ -90,12 +90,13 @@ export default class FormSelect3 extends AFormElement<IFormSelect3> {
    */
   set value(v: (ISelect3Item<IdTextPair> | string) | (ISelect3Item<IdTextPair> | string)[]) {
     const toIdTextPair = (d) => {
-      if (!d.id && !d.text) {
-        return {id: null, text: null};
-      }
-      return {
-        id: d.id ? d.id : d.text,
-        text: d.text ? d.text : d.id
+      if (typeof d === 'string') {
+        return {id: d, text: d};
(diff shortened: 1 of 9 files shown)


src/form/internal/Select3.ts
```diff
@@ -0,0 +1,489 @@
...
+
+/**
+ * Replacer function that styles the found match, offset 0 means no match
+ * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace
+ * @param match The matched substring
+ * @param p1 The nth parenthesized submatch string
+ * @param offset The offset of the matched substring
+ * @returns {string} The replacement string
+ */
+export function highlightMatch(match, p1, offset) {
```
sgratzl: missing typings

src/form/internal/FormSelect3.ts
```diff
@@ -0,0 +1,108 @@
...
+      return v !== '' || (<any>v).id !== '';
+    }
+  }
+
+  /**
+   * Select the option by value. If no value found, then the first option is selected.
+   * @param v If string then compares to the option value property. Otherwise compares the object reference.
+   */
+  set value(v: (ISelect3Item<IdTextPair> | string) | (ISelect3Item<IdTextPair> | string)[]) {
+    if (Array.isArray(v) && v.length > 0 && !this.multiple) {
```
sgratzl: since you can return just the id or text, it is also possible just to set them. Thus, the value can be 
 * an array of items
 * an array of string (id or text) depending on this.options.return
 * an item (!multiple)
 * a string (!multiple & this.option.return...)

PR #92, opened 2018-03-08
fix #90 support table view sorting

### LINKED ISSUE #90: support sorting and ATableView
to improve the ATableView simple operations such as sorting would be useful.

sgratzl:
fix #90


sluger:
as discussed, a sort order over multiple columns is not part of the issue.

PR #104, opened 2018-03-28
use decimal encoder

lehnerchristian:


lehnerchristian:
@sgratzl , I've updated this PR to use the `pandas.json`'s `double_precision` parameter as stated in https://github.com/phovea/phovea_server/pull/64#issuecomment-376876439

this is only used when sending the request to the view with the `.json` suffix

PR #105, opened 2018-04-11
enable export when the table is rendered

lehnerchristian:
otherwise the export is only enabled whenever the table is instantiated, but overidden when it's re-rendered


sluger:
thanks

PR #107, opened 2018-04-23
Content generator

sgratzl:
requires https://github.com/phovea/phovea_ui/pull/52

add options in ITDPOptions for advanced header


src/styles/_datavisyn.scss
```diff
@@ -48,6 +48,7 @@ nav.navbar-inverse {
 
 .#{$datavisyn_css_prefix}-font {
   @include datavisyn-font();
+  line-height: 1.42857;
```
sluger: can you elaborate on this magic number please?

PR #110, opened 2018-06-14
Sgratzl/form initial value

### LINKED ISSUE #7: use previous session settings are not stored in provenance graph
when replaying actions can be compressed, e.g. certain views never shown since they will be replaced anyhow this includes executing all the parameter settings. 

However, with the "useSession" feature in which previous filter settings are stored, these stored values will never be set and the initial parameters are not stored. 

possible solution: 
 * when initializing a form builder element with a session value it should also trigger the change event which will trigger a recording of the parameter change

sgratzl:
fix #7

workaround for storing the view parameters that are initialized based on the current user session, i.e. based on other form fields with the same name.


lehnerchristian:
how can I test this PR? not sure what steps I need to follow to reproduce the problem correctly

lehnerchristian:
I've tested this issue and found that it generally works, but the following scenario didn't work:
* load a list of all genes
* select a gene
* select the Copy Number detail view
* change the data source parameter to Tissue
* swich to the Expression view
* clear the session storage
* reload (everything works fine until here=
* after reloading, switch back to the Copy Number detail view --> notice that `Cell Line` is selected again
* switch back to Expression --> Cell Line is selected here, too

tdp_core_parameters

sgratzl:
please test again

lehnerchristian:
works now. the code also looks good from my point of view (therefore I'd approve). however if it doesn't need to be merged as soon as possible I'd test it again on monday

lehnerchristian:
When saving a session persistently and loading it again from the start menu, I've the following JS error:
parametertest

The problem does not occur, when I clear the sessionStorage before loading the session. But the second time loading the session leads back to the error.

when I check the normal lineupjs3 branch out, I get a JS error `Type error: n is null`, however restoring the view works

sgratzl:
what is the current status? what was the session. how can I reproduce the bug if needed?

lehnerchristian:
Ok, I've tested again

1. parameter gets reset when I clear the session storage and don't reload
* Select a list of all genes
* Select a gene
* Select the Copy Number detail view
* Change the parameter to Tissue
* Switch to the Expression detail view
* Clear the session storage
* Switch back to the CN detail view
* The parameter is reset

2. I can reproduce the `TypeError: this.$select is undefined` issue without storing the Provenance Graph permanently (as shown [above](https://github.com/datavisyn/tdp_core/pull/110#issuecomment-397982537))
* Open a list of all genes
* Open the Combined View
* Select a data subtype (e.g. Relative Copy Number)
* Reload
* The error should be shown in the console
* I cannot reproduce this in the normal `tdp_core#lineupjs3` branch

sgratzl:
> parameter gets reset when I clear the session storage and don't reload

what is the expected outcome when you delete the stored value while the application is running?

lehnerchristian:
nothing, it's just that you told me some weeks ago that I should test it by clearing the session storage. but I guess it's the normal behavior, since the parameters seem to be stored in the sessionStorage

PR #112, opened 2018-06-23
Swagger

sgratzl:
improve swagger generation


tdp_core/swagger/swagger.yml
```diff
@@ -2,7 +2,7 @@ openapi: 3.0.0
 servers:
   - url: /api/tdp/
 info:
-  version: 4.0.1
+  version: 4.0.2
```
sluger: we have to set the version number automatically for future releases

sluger:
please let me know if this can be merged already, or just do it on your own, thanks

PR #118, opened 2018-07-30
Sgratzl/lineupjs3.1

sgratzl:
updates to lineupjs3.1 

"should" also work with lineupjs3.0 (but not tested)

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@types/d3": "3.5.36",
     "@types/select2": "4.0.41",
     "d3": "3.5.17",
-    "lineupjs": "~3.0.1-beta.11",
+    "lineupjs": "~3.1.0",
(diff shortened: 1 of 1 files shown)


sluger:
should this work with patentbay too or are there some adjustments to make before using lineupjs@3.1?

lehnerchristian:
looks good in general, also the sidebar styling issue is resolved
however, as discussed yesterday, on my system I get scrollbars in most of the cells and there is the font rendering issue in score columns when the bar and the text overlap
peek 2018-08-01 08-47

sgratzl:
what happens on https://lineup.js.org/app

sgratzl:
and regarding the coloring bug -> https://github.com/datavisyn/lineupjs/issues/47

lehnerchristian:
no scrollbars:

peek 2018-08-01 08-54

PR #119, opened 2018-07-31
move swagger path description and summary for generic views into http…

sluger:
… method

The swagger UI did not show path descriptions and summaries as the json properties were appended to the path instead of the http method level. This PR fixes that by moving it one level down, e.g in yaml format: 

**current**:
```yaml
/generic/view/path:
    summary: 'some summary'
    description: 'any description'
    get:
      ...
```

**fix**:
```yaml
/generic/view/path:
    get:
      summary: 'some summary'
      description: 'any description'
      ...
```


sgratzl:
please create a patch for the `develop` branch that is currently being used in all the deployed versions

PR #133, opened 2018-11-12
Clehner/custom column colors

lehnerchristian:
I added some minor changes in order to set specific colors to instead of letting the color manager do all the work (there was also a `color` attribute in the basecolumn, which could not be set)

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -45,7 +45,7 @@ export interface IScore<T> {
    * creates the LineUp column description
    * @returns {IColumnDesc & {[p: string]: any}}
    */
-  createDesc(): IColumnDesc & {[key: string]: any};
+  createDesc(extras?: object): IColumnDesc & {[key: string]: any};
 
 
   /**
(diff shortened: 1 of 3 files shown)


sgratzl:
The `color` attribute of a description is going to be deprecated in the next version. Instead for numerical columns, please use the `colorMapping` attribute. This allows a more complex color mappings (gradients, custom mapping, ...) see https://github.com/datavisyn/lineupjs/blob/7b98b6e24b7bdf09a874a9c6245a09398960b215/src/model/MappingFunction.ts#L251

lehnerchristian:
so I need to rename the attribute `color` in https://github.com/datavisyn/tdp_core/blob/faf268380a1a91bdfbd3d7d0f31bd3f173bc8b87/src/lineup/desc.ts#L51-L55 to `colorMapping` and use that one instead?

lehnerchristian:
oh this is only valid for numerical columns. so instead I'd rather set the option in the `numberCol` factory?

i.e.:
https://github.com/datavisyn/tdp_core/blob/clehner/custom_column_colors/src/lineup/desc.ts#L83:L96

lehnerchristian:
I removed the color option again, since the new `colorMapping` is can be passed as an `extra`

PR #134, opened 2018-11-13
Sgratzl/tour

sgratzl:
implement the basis for generating TDP tours. inspired by https://introjs.com/ but goes way beyond that. 

A tour is described as an extension, see 

https://github.com/datavisyn/tdp_core/blob/67dbb29bb7e4a665a3c69386422448ed1a13ff41/src/tour/extensions.ts#L9-L76

consisting of multiple steps. Each step can optionally highlight an element and include custom `preAction`, `postAction`, `waitFor` callbacks


thinkh:
@sgratzl Nice work. 👍 Where did you configure/implement that you presented in [this comment](https://github.com/Caleydo/tdp_bi_bioinfodb/issues/513#issuecomment-438243645)

sgratzl:
> @sgratzl Nice work. 👍 Where did you configure/implement that you presented in [this comment](https://github.com/Caleydo/tdp_bi_bioinfodb/issues/513#issuecomment-438243645)?

since the tour is a combination of ordino and tdp_publicdb -> tdp_publicdb with a restriction for ordino only

PR #146, opened 2018-12-03
add option to show side panel in the collapsed version in the top row…

sgratzl:
… instead of side. Disables side panel for this and just shows the buttons

2018-12-03 17_51_03-patentbay


sgratzl:
actually just moving around some rules + adding the overrides for this new side panel model. Seems like that Git wasn't smart enough to proper detect that

PR #149, opened 2019-01-07
Sgratzl/advancedsec

sgratzl:
implement a common permission form component to edit advanced permissions for an item. 

e.g. during persisting the provenance graph

image

### DIFF
--- a/src/ATDPApplication.ts
+++ b/src/ATDPApplication.ts
@@ -241,18 +241,17 @@ export abstract class ATDPApplication<T> extends ACLUEWrapper {
     if (plugins.length === 0) {
       return;
     }
-    Promise.all([<any>this.app, ...plugins.map((d) => d.load())]).then((args) => {
-      const appInstance = args[0];
-      const plugins: IAppExtensionExtension[] = args.slice(1);
-
-      for (const plugin of plugins) {
-        plugin.factory({
-          header: this.header,
-          content,
-          main,
-          app: appInstance
-        });
-      }
+    this.app.then((app) => {
+      Promise.all(plugins.map((d) => d.load())).then((plugins: IAppExtensionExtension[]) => {
+        for (const plugin of plugins) {
+          plugin.factory({
+            header: this.header,
+            content,
+            main,
+            app
+          });
+        }
+      });
     });
   }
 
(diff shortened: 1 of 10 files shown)


thinkh:
I find this representation a bit confusing.

In the screenshot the *Private* radio button and the *Advanced* button is selected. In my opinion it doesn't make sense to change some settings for public, groups, and buddies, when my session is private.

Hence, I suggest to show/activate the *Advanced* button only when the radio button *Public* is selected. When *Private* is selected the *Advanced* button should be disabled or hidden.

Further, I suggest to rename the label *Private* in the blue toggle buttons to *Forbidden* or *No permission*. Then the buttons get clearer, since a group can have *no permission*, *read permission*, or *write permission*. Otherwise mix between public (group label) and private (from the toggle button) is confusing.

sgratzl:
> In the screenshot the _Private_ radio button and the _Advanced_ button is selected. In my opinion it doesn't make sense to change some settings for public, groups, and buddies, when my session is private.

re advanced -> by default it just shows the simplified public/private version. upon click on advanced you will get the advanced options. 

> 
> Hence, I suggest to show/activate the _Advanced_ button only when the radio button _Public_ is selected. When _Private_ is selected the _Advanced_ button should be disabled or hidden.

public / private is synced with the Public option. Thus even if it is private for others it might be still writeable for some buddies

> 
> Further, I suggest to rename the label _Private_ in the blue toggle buttons to _Forbidden_ or _No permission_. Then the buttons get clearer, since a group can have _no permission_, _read permission_, or _write permission_. Otherwise mix between public (group label) and private (from the toggle button) is confusing.

done

thinkh:
Thanks for the clarification and the change.

If I understood you correctly, the radio button at the top influences only the (first) public group.

* Select radio button *private* == *no permission* toggle button of the public group 
* Select radio button *public* == *write* toggle button of the public group 

Which radio button will be selected if the *read* toggle button of the public group is selected?

Maybe it would be better to remove the public group (under advanced) and add another radio button at the top instead. The final result would be three radio buttons:

* Private
* Public (read-only)
* Public (read/write)

What do you think?

sgratzl:
> * Select radio button _public_ == select _write_ toggle button of the public group

no it is just read access. 

I don't want to make the default security view

image

more difficult. just upon clicking advanced you have more options. The private / public metaphor is a very generalized view. The advanced options are more granular. e.g. I would not consider it public if I just allow a buddy to read it.

PR #152, opened 2019-02-04
Sgratzl/base

sgratzl:
move some sqlalchemy engine related options to the connector for easier subclassing and customizing

add a base helper class to add an embedded ranking view


src/base/AEmbeddedRanking.ts
```diff
@@ -0,0 +1,139 @@
...
+      }
+
+      runWithoutTracking<T>(f: () => T): Promise<T> {
+        return super.withoutTracking(f);
+      }
+    }
+
+    this.ranking = new EmbeddedRankingView(context, selection, this.node, options);
+
+    // since set in the constructor it is safe
```
lehnerchristian: what to you mean with safe? when would the assignment in the next line be not safe?

src/base/AEmbeddedRanking.ts
```diff
@@ -0,0 +1,139 @@
...
+    const lineup = this.data;
+    lineup.on(LocalDataProvider.EVENT_SELECTION_CHANGED + '.embedded', (selection: number[]) => {
+      const rows = selection.map((d) => lineup.data[d]);
+      this.selectedRowsChanged(rows);
+    });
+
+    const dummy = this.node.ownerDocument.createElement('div');
+    this.ranking.init(dummy, () => null);
```
lehnerchristian: why do you need this dummy?

PR #157, opened 2019-02-13
Clehner/wip fix tour

lehnerchristian:
the tour now considers the documents full height (including the scroll bar) and the scroll offset (because `mask.top` varies depending on the scroll offset)

### DIFF
--- a/src/tour/TourManager.ts
+++ b/src/tour/TourManager.ts
@@ -171,15 +171,18 @@ export default class TourManager {
     this.backdrop.style.height = `${fullAppHeight}px`;
     this.backdrop.style.width = `${fullAppWidth}px`;
 
+    // also consider the current scroll offset inside the window
+    const scrollOffset = self.scrollY;
+
     // @see http://bennettfeely.com/clippy/ -> select `Frame` example
     this.backdrop.style.clipPath = `polygon(
       0% 0%,
       0% ${fullAppHeight}px,
       ${mask.left}px ${fullAppHeight}px,
-      ${mask.left}px ${mask.top}px,
-      ${mask.left + mask.width}px ${mask.top}px,
-      ${mask.left + mask.width}px ${mask.top + mask.height}px,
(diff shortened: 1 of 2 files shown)


lehnerchristian:
@sgratzl , can you please have a look at this small PR? I need to test it in the deployed version, if it's ok

src/tour/TourManager.ts
```diff
@@ -164,18 +164,28 @@ export default class TourManager {
...
   private setHighlight(mask: IBoundingBox) {
+    const fullAppHeight = this.backdrop.ownerDocument.body.scrollHeight; // full page height including scroll bars
+    const fullAppWidth = this.backdrop.ownerDocument.body.scrollWidth; // full page width including scroll bars
+
+    // set the new height of the backdrop
+    this.backdrop.style.height = `${fullAppHeight}px`;
+    this.backdrop.style.width = `${fullAppWidth}px`;
+
+    // also consider the current scroll offset inside the window
+    const scrollOffset = self.scrollY;
```
sgratzl: `scrollY` is a property of which object?

src/tour/TourManager.ts
```diff
@@ -164,18 +164,28 @@ export default class TourManager {
...
-      ${mask.left}px ${mask.top}px,
-      ${mask.left + mask.width}px ${mask.top}px,
-      ${mask.left + mask.width}px ${mask.top + mask.height}px,
-      ${mask.left}px ${mask.top + mask.height}px,
-      ${mask.left}px 100%,
-      100% 100%,
-      100% 0%
+      0% ${fullAppHeight}px,
+      ${mask.left}px ${fullAppHeight}px,
+      ${mask.left}px ${mask.top + scrollOffset}px,
```
sgratzl: what happens when the user scrolls? or why is the scroll offset not considered when computing the mask?

src/tour/TourManager.ts
lehnerchristian: the window

PR #161, opened 2019-02-21
Clehner/html tooltip

lehnerchristian:


src/views/tooltip.ts
```diff
@@ -92,7 +98,7 @@ function isRelated(evt: MouseEvent, ref: HTMLElement) {
  * @param {HTMLElement} reference
  * @param { number, number } coords
  */
-export function popOver(html: string | (() => string), reference: HTMLElement, coords?: { x: number, y: number }, simpleTooltip: boolean = true) {
+export function popOver(html: string | HTMLElement | (() => string), reference: HTMLElement, coords?: { x: number, y: number }, simpleTooltip: boolean = true) {
```
sgratzl: for the sake of completeness: `html: string | HTMLElement | (() => string | HTMLElement)` i.e the generator can also produce HTML

PR #168, opened 2019-05-20
correct computation of more

lehnerchristian:


thinkh:
Please have a look at `limit_offset` from https://github.com/datavisyn/tdp_core/blob/dffc8df49cf7cd05c94e5c50af181c2308aa572c/tdp_core/dbview.py#L367.

PR #171, opened 2019-06-28
add list as tuple param

anita-steiner:
* add a new param: list_as_tuple
this will generate a string of tuples (1,<element1>), (1,<element2), ....
this is especially for oracle because of the 1000 elements limit for IN statements


anita-steiner:
@lehnerchristian nur falls du noch dazu kommst

Issue #181, opened 2019-08-06
Add idType to db view description


keckelt:
When querying a view description from the server, it returns the columns and the idType:
https://github.com/datavisyn/tdp_core/blob/ce8a40ff1b3a2c0b55d1a107ad79063f4a737ce6/tdp_core/sql.py#L146-L149

On client-side however, the interface only defines the columns as property:
https://github.com/datavisyn/tdp_core/blob/35b3e62973ba3fd07d50d92902dfeaac992cbc51/src/rest.ts#L205-L210

The idType should added to the interface as the server already returns it.

Note: returned idType can be `null`, e.g. for views `cellline_unique_all`, `cellline_check_ids`, `cellline_panel`.

keckelt:
The PhoveaDataAdapter also creates a ViewDescription:
https://github.com/datavisyn/tdp_core/blob/ce8a40ff1b3a2c0b55d1a107ad79063f4a737ce6/src/internal/PhoveaDataAdapter.ts#L16-L24

It is created from an ITable, which also stores the idType so i can use that for the view description.

keckelt:
The `loadColumnDesc` function in `AEmbeddedRanking` also creates a ViewDescription that had to be extended.

PR #182, opened 2019-08-06
Refactor DB view descriptions

keckelt:
Please see #181


thinkh:
We agreed in a personal meeting for the following changes:

* Rename `IViewDesc` to `IDBViewDesc`
* Unify return values for API routes `/tdp/db/publicdb/` and `/tdp/db/publicdb/myview/desc` so that they match the `IDBViewDesc`
* `IViewDescription` is only used by the `PhoveaDataAdapter` and should be moved there
* `ARankingView` and `AEmbeddedRankingView` should use the `IDBViewDesc` instead

@keckelt Please change the code accordingly and notify @lehnerchristian. Afterwards he will test the changes in different apps.

keckelt:
@lehnerchristian @thinkh I changed the points we discussed, except:
> * IViewDescription is only used by the PhoveaDataAdapter and should be moved there
> * ARankingView and AEmbeddedRankingView should use the IDatabaseViewDesc instead

The RankingViews only use the interface to get the column description,so using the `IDatabaseViewDesc` is an overkill.
Actually it might be better to just return `IServerColumn[]` there, but I left it untouched for now as that might be best handled in another PR imo.

src/rest.ts
```diff
@@ -39,18 +39,29 @@ export function getTDPDatabases(): Promise<IDatabaseDesc[]> {
   return getAPIJSON(`${REST_DB_NAMESPACE}/`);
 }
 
-export interface IViewDesc {
+export interface IViewDescription {
+
+  /**
+   * idType of the DBView, can be null
+   */
+  idType: string | null;
```
lehnerchristian: I think we should also add the type `IDType` here, so it should be `idType: IDType | string | null;`. the reason for this is that we might also want to use the actual resolved IDType in the frontend and not only the string (e.g. `GeneSymbol_human`).

src/rest.ts
```diff
@@ -39,18 +39,29 @@ export function getTDPDatabases(): Promise<IDatabaseDesc[]> {
...
+
+  /**
+   * list of columns within this view, can be empty
+   */
+  columns: IServerColumn[];
+}
+
+export interface IDatabaseViewDesc extends IViewDescription {
```
lehnerchristian: should we make the attributes in `IViewDescription` optional again like it was in the original version? not sure if we catch all potential problems in each plugin. making them optional again should be backwards compatible. if so it would be `... extends Partial<IViewDescription>`

src/rest.ts
lehnerchristian: or the shorter version would be `idType: IDTypeLike | null` --> and I'm not sure if we need the `| null`

lehnerchristian:
TODO after merging: also merge dependent PRs where typings change

src/rest.ts
keckelt: it's only needed if you have `--strictNullChecks`, but I thought I include it because I already know it can be null.

keckelt: Set it to `idType: IDTypeLike | null`.

src/rest.ts
keckelt: Yes, updated the definition.

keckelt:
Ah, jetzt wo ich die member von IViewDescription optional gemacht habe ist natürlich IDatabaseViewDesc keine richtige Erweiterung:
```
Type 'Promise<Readonly<IDatabaseViewDesc>>' is not assignable to type 'Promise<IViewDescription>'.
  Type 'Readonly<IDatabaseViewDesc>' is not assignable to type 'IViewDescription'.
    Property 'idType' is optional in type 'Readonly<IDatabaseViewDesc>' but required in type 'IViewDescription'.

```

lehnerchristian:
in which project do you get the error? I've checked your branch out in my local version of Ordino and I don't get any compile errors

keckelt:
In Ordino, I have errors in the subclasses of ARankingView like: https://github.com/Caleydo/tdp_publicdb/blob/master/src/views/ACombinedDependentTable.ts#L77

I think we should change the [return type of `loadColumnDesc`](https://github.com/datavisyn/tdp_core/blob/master/src/lineup/ARankingView.ts#L465), but don't know how much work that would be (e.g. to `IServerCoumn[]`), or if that is better in another PR.

keckelt:
Return of `/api/tdp/db/<db>/<view>/desc` before and after 
Old:
```json
{
  "idType": null,
  "columns": []
}
```

New: 
```json
{
  "description": "",
  "arguments": [
    "query",
    "query",
    "species",
    "column",
    "limit",
    "offset"
  ],
  "query": "SELECT s as id, s as text FROM (SELECT distinct {column} AS s FROM tissue.tdp_tissue WHERE LOWER({column}) LIKE :query AND species = :species) ORDER BY {column} ASC LIMIT {limit} OFFSET {offset}",
  "type": "helper",
  "columns": [],
  "name": "tissue_unique"
}
```

keckelt:
Hi @lehnerchristian, I decided to not make the `IViewDescription` properties optional because they are always returned by the  API (idType can be null/the array can be empty though).

lehnerchristian:
ok, thank you. I'll have a look

lehnerchristian:
sorry, I'll need to have a look at all projects, because this PR has some breaking changes. I talked to @thinkh and we're checking how we release it: a new major version + probably renaming of IViewDescription to IViewDesc

lehnerchristian:
I've checked this PR with several projects especially if there are some RankingViews returning only columns without IDTypes . I'll need to check some other projects, which I'm not involved in with some colleagues tomorrow and then I can hopefully merge this PR

src/rest.ts
```diff
@@ -39,18 +40,29 @@ export function getTDPDatabases(): Promise<IDatabaseDesc[]> {
   return getAPIJSON(`${REST_DB_NAMESPACE}/`);
 }
 
-export interface IViewDesc {
+export interface IViewDescription {
```
thinkh:
```suggestion
export interface IViewColumnDesc {
```

src/rest.ts
```diff
@@ -39,18 +40,29 @@ export function getTDPDatabases(): Promise<IDatabaseDesc[]> {
...
+   */
+  idType: IDTypeLike | null;
+
+  /**
+   * list of columns within this view, can be empty
+   */
+  columns: IServerColumn[];
+}
+
+export interface IDatabaseViewDesc extends IViewDescription {
```
thinkh:
```suggestion
export interface IDatabaseViewDesc extends IViewColumnDesc {
```

lehnerchristian:
@dg-datavisyn will test this with his projects, too and then this change will be in the next release in 2 weeks

in the meantime we'll merge this into develop after the release we'll create now.

thank you for the changes and sorry that it took me so long, @keckelt :wink:

src/rest.ts
keckelt: Sounds more like it is a description of the views columns, imo.

thinkh: Mh, true. Our rationale was to unify it to `...Desc`, which was not possible in this case, because `IViewDesc` is already in use. So we came up with this interface name. Do you have a different suggestion?

lehnerchristian: the rationale behind the naming @thinkh proposed was that we wanted to rename it to have the `Desc` suffix that it is consistent with the other names. and we cannot name it IViewDesc, beacuse that was the original name of `IDatabaseViewDesc`

I'd say we either name it `IServerColumnDesc`, because it contains an `IServerColumn[]` + and IDType as meta data

keckelt: Now that I think about it, I guess both is fine (`IViewColumnDesc`/`IServerColumnDesc`)  because the interface is no longer used to describe a view but a set of columns.

thinkh: Then let's go for `IServerColumnDesc`. Could you please change this?

lehnerchristian:
as soon as it's renamed as @thinkh suggested we'll merge it

lehnerchristian:
@keckelt , can you rename the interfaces as @thinkh suggested?

src/rest.ts
keckelt: Done

keckelt:
@lehnerchristian Done :slightly_smiling_face:

PR #185, opened 2019-08-06
Allow custom form elements with extension point `epTdpCoreFormElement`

### LINKED ISSUE #192: Add custom form elements to dialog
I want to extend the dialog with custom form elements that are not implemented/provided by tdp_core.

image

Possible solutions:

1. Bypass the form builder and set "raw" HTML to the dialog
2. Extend form builder to accept custom form elements

Solution 1 is very hard-coded and not flexible for further extensions. Solution 2 has the advantage that new form elements can be provided by different plugins/repos that are not part of tdp_core.

thinkh:
Fixes #192

This PR introduces the new extension point `epTdpCoreFormElement` to register and use custom form elements.

**Related PRs**

* https://github.com/Caleydo/tdp_gene/pull/98
* https://github.com/Caleydo/tdp_publicdb/pull/51

## Breaking changes

* **Change:** `AView.init()` returns (officially) a promises
  **Fix:** Adapt to the new return value and return a Promise (or empty Promise)
  ```ts
  export abstract class MyCustomView extends AView {

    init(params: HTMLElement, onParameterChange: (name: string, value: any, previousValue: any) => Promise<any>): Promise<any> { // update to new return type
      // ...
      return this.waitForInit(); // if there is something to wait, return the promise
      // OR
      return Promise.resolve(); // if there is nothing to wait, resolve an empty promise
    }
  }
  ```
* **Change:** The form elements are moved from `form/internal` to `form/elements`
  **Fix:** Update the imports in TypeScript, as for instance:
   
  ```diff
  -import FormSelect3 from 'tdp_core/src/form/internal/FormSelect3';
  +import FormSelect3 from 'tdp_core/src/form/elements/FormSelect3';
  ```
* **Change:** `FormBuilder.build()` returns an `Promise<IForm>` (previously named `IParent` and synchronous)
  **Fix:** Wait for the Promise to resolve and use the _IForm_ as the FormBuilder before
  ```ts
  // old synchronous behavior
  const builder = new FormBuilder($parent);
  builder.build(elementDescs);
  builder.getElementById('...'); // any FormBuilder function

  // new asynchronous behavior
  const builder = new FormBuilder($parent);
  builder.build(elementDescs).then((form) => {
    form.getElementById('...'); // any IForm function
  }
  ```

## Other changes

* Add phovea extension point `epTdpCoreFormElement`
* Register currently available form elements as phovea extension
* The `create` function for form elements returns a promise, due to async loading of the phovea plugins
* The elements of the enum `FormElementType` needs to be string (otherwise the form elements will be registered as integer number in _phovea.js_)
* The form elements are moved from `form/internal` to `form/elements`

Further changes due to handling dependent form elements:

* Separate `Form` from `FormBuilder`: now the `FormBuilder` returns an `IForm`
   * Rename `IFormParent` to `IForm` and list missing methods
* Add `init()` to `IFormElements`
   * All elements of the form are available at initialization time
   * Supports dependencies to other elements (i.e., `handleDependent()`)
* `FormBuilder.build()` is async
   * Reason: We must wait for all elements to be loaded and initialized
* `AView.init()` returns (officially) a promises
   * before it was typed as any, but returned a promise
   * change was necessary due to the async `FormBuilder.build()`

## How to add custom form elements

Register the custom form element in _phovea.js_ of your plugin:

```js
registry.push('epTdpCoreFormElement', 'CustomInputTextID', function () {
   return import('./src/form/FormCustomInputText');
});
```

Implement the element in _src/form/FormCustomInputText.ts_ by implementing the `IFormElement` interface (or extending `AFormElement`):

```ts
import * as d3 from 'd3';
import AFormElement from './AFormElement';
import {IFormElementDesc, IFormParent} from '../interfaces';

/**
 * Add specific options for input form elements
 */
export interface IFormCustomInputTextDesc extends IFormElementDesc {
  /**
   * Additional options
   */
  options?: {};
}

export default class FormCustomInputText extends AFormElement<IFormInputTextDesc> {

  private $input: d3.Selection<any>;

  /**
   * Constructor
   * @param form The form this element is a part of
   * @param $parent The parent node this element will be attached to
   * @param elementDesc The form element description
   * @param pluginDesc The phovea extension point description
   */
  constructor(parent: IFormParent, $parent, elementDesc: IFormInputTextDesc, readonly pluginDesc: IPluginDesc) {
    super(parent, desc, pluginDesc);

    this.$node = $parent.append('div').classed('form-group', true);

    this.build();
  }

  /**
   * Build the label and input element
   * Bind the change listener and propagate the selection by firing a change event
   */
  protected build() {
    super.build();
    this.$input = this.$node.append('input').attr('type', 'text');
  }

  /**
   * Returns the value
   * @returns {string}
   */
  get value() {
    return this.$input.property('value');
  }

  /**
   * Sets the value
   * @param v
   */
  set value(v: string) {
    this.$input.property('value', v);
    this.previousValue = v; // force old value change
    this.updateStoredValue();
  }

  focus() {
    (<HTMLInputElement>this.$input.node()).focus();
  }
}
```

Use custom element for instance in a score dialog:

```ts
  const dialog = new FormDialog('Add custom score', 'Add score');
  const custom: IFormElementDesc = {
    type: 'MyCustomFormElementID', // use the extension id from phovea.js
    label: 'My custom element...',
    id: 'my-custom',
    options: {}
  };
  dialog.append(custom);
```

## Open for discussion

1. Since all TDP form elements are available as extension, it does not make sense that the TS files are located in the directory `/form/internal`. Should we rename the directory? The alternative would be to keep the TDP form elements hidden and do not expose them as extension.
   * We decided to rename the directory `/form/internal` to `/form/elements`
2. At the moment the [create function handles a special case](https://github.com/datavisyn/tdp_core/pull/185/files#diff-d78b2dbe3c77ad0d08c5cf1668472da0R25-R27) for the selection option for Select2 and Select3; should we find a more general approach for all kind of options from the phovea extension? Or do we keep it like this until we need a more general solution?
   * We decided to pass the `pluginDesc` by default to the form element and each form element can decided how to handle the phovea extension options


thinkh:
Open for discussion:

1. Since all TDP form elements are available as extension, it does not make sense that the TS files are located in the directory `/form/internal`. Should we rename the directory?
2. At the moment the [create function handles a special case](https://github.com/datavisyn/tdp_core/pull/185/files#diff-d78b2dbe3c77ad0d08c5cf1668472da0R25-R27) for the selection option for Select2 and Select3; should we find a more general approach for all kind of options from the phovea extension?

thinkh:
There is still a bug when using dependent form elements, as some elements might not be available when initializing the dependent elements.

image

image

thinkh:
A bunch of changes in the form building process where necessary, in order to fix the handling of dependent elements:

* Separate `Form` from `FormBuilder`: now the `FormBuilder` returns an `IForm`
   * Rename `IFormParent` to `IForm` and list missing methods
* Add `initialization()` to `IFormElements`
   * All elements of the form are available at initialization time
   * Supports dependencies to other elements (i.e., `handleDependent()`)
* `FormBuilder.build()` is async
   * Reason: We must wait for all elements to be loaded and initialized
* `AView.init()` returns (officially) a promises
   * before it was typed as any, but returned a promise
   * change was necessary due to the async `FormBuilder.build()`

thinkh:
* Move files from `form/internal/` to `form/elements/`
* Add list of breaking changes to this PR
* Use [plugin.get](https://github.com/phovea/phovea_core/blob/master/src/plugin.ts#L174-L176) in `create` function
* Pass phovea extension option by default to form element (in `create` function)

thinkh:
@lehnerchristian I addressed the missing point, discussed in the last code review meeting, with my most recent commits. You can start reviewing. 😃

thinkh:
@lehnerchristian I updated the code according to our extension point convention:

* Define extension point as constant in _extension.ts_
* Add documentation

lehnerchristian:
works perfectly in my opinion :+1: 
I've tested this with various applications

only 1 thing I discovered: I think we probably should not move the file `Select3.ts` to `tdp_core/src/form/elements` but rather leave it in `tdp_core/src/form/internal` because some applications import the Select3 class directly and then things break

Issue #19, opened 2017-09-15
refactor formbuilder


sgratzl:
the form builder is the only piece left in the core using d3. I order to avoid using d3(v3) in the core the FormBuilder needs to be refactored. 

At the same time it can be improved. 
 * Atm there is no overview what are the options for specific form element descs. A builder pattern could help in this case.
 * similarly when calling `getParameter` the view has to now which are valid parameter and what will be the return value. When using an interface object (IParams...) everything could be typed. 
 * select vs select2 both have different styles how to provide data, name/value vs id/text in which forms (data, promise, ajax, ... This should be refactored
 * simplify select2 similarly to the search box in the Dashboard application, i.e. implementing a custom transport method
 * provide option to extend a form desc e.g. with a visualization to generate the parameter like a histogram selector

sgratzl:
related #25

thinkh:
It's also related to #52

lehnerchristian:
we could either do this by using union types:

```
interface ISelectDesc {
  type: 'select';
  name: string;
}
interface ITextDesc {
  type: 'text';
  value: string;
}

declare type IDesc = ISelectDesc | ITextDesc;

const t: IDesc[] = [
  {
    type: "select",
    name: 'test'
  },
  {
    type: 'text',
    value: 'a'
  }
];
```

or with the builder pattern as a wrapper around the usual form description objects (i.e. the `buildSelectElement` function returns an object with the correct form element type):
```
buildSelectElement(ParameterFormIds.DATA_SOURCE).label('Data Source').required().useSession().optionsData(..).build()
```

PR #186, opened 2019-08-07
Remove D3 from form builder

thinkh:
Part of #19

_FormRadio_ and _FormSelect_ are the only classes left that use a D3 data join.

grafik

Overall there are more classes using D3 modules:

grafik

### DIFF
--- a/src/form/FormBuilder.ts
+++ b/src/form/FormBuilder.ts
@@ -2,7 +2,6 @@
  * Created by Samuel Gratzl on 08.03.2017.
  */
 
-import * as d3 from 'd3';
 import {randomId} from 'phovea_core/src/index';
 import {IFormElement, IFormElementDesc, IForm} from './interfaces';
 import {create} from './internal';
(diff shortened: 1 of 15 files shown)


thinkh:
Tip: Could use [insertAdjacentHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML) to append new HTML elements.

thinkh:
Currently there is still a bug with the call of `showIf` for dependent elements.

1. Open any gene list
2. Add _Aggregated Score Column_
3. Select _Data Type: TPM_ and _Aggregation: Frequency_

It should show the drop-down elements _Comparison Operator_ and _Comparison Value_.

I assume that the event listener to the change event is not correctly added in `AFormElement.handleDependent()`.

lehnerchristian:
I've changed the base branch to develop, because the other branch was merged

thinkh:
I close this PR. The removal of D3 has no priority and the usage of the form builder will fade-out. Over time it will be replaced by a [React form library](https://github.com/datavisyn/visyn2025/issues/132).

PR #190, opened 2019-08-14
added new logo

dvvanessastoiber:
New Logo in the Main View of Jamborees:
Screenshot from 2019-08-16 06-31-05

New Logo when opening a Jamboree:
Screenshot_openedJamboree

This change of the logo can only be done in tdp_core.


mstreit:
@dvvanessastoiber, can you please add a screenshot to PRs that contain visual changes? that makes it easier for me and other non developers to provide feedback. thanks!

dvvanessastoiber:
> @dvvanessastoiber, can you please add a screenshot to PRs that contain visual changes? that makes it easier for me and other non developers to provide feedback. thanks!

Of course, I'm sorry!

thinkh:
Could we either make the d-logo a bit smaller in height or increase the font-size of the Jamborees title? Because like this it looks unbalanced.

dvvanessastoiber:
> Could we either make the d-logo a bit smaller in height or increase the font-size of the Jamborees title? Because like this it looks unbalanced.

Is it better like that?
Screenshot_resize(1)
Screenshot_resize

mstreit:
In the screen design Ana made the logo the same size as the app name: 
image

We could also have the app name in the same style as our logo (with colored dots, if present):
image

lehnerchristian:
I also think that the logo and the name of the application should be of the same size. can we try that out?

dvvanessastoiber:
Screenshot_resized

Would that be okay?

lehnerchristian:
looks good in my opinion :+1: 
@mstreit and @thinkh , what would you say?

mstreit:
yep, looks good!

src/styles/_datavisyn.scss
```diff
@@ -28,24 +28,33 @@ $dataviyn_font: Monda !default;
 
 nav.navbar,
 nav.navbar-inverse {
+  min-height: auto; // Temp fix until https://github.com/phovea/phovea_clue/issues/140 is fixed
```
thinkh: It is fixed with https://github.com/phovea/phovea_ui/pull/69 and already merged into develop.

src/styles/_datavisyn.scss
```diff
@@ -28,24 +28,33 @@ $dataviyn_font: Monda !default;
...
     &:before {
       @include datavisyn-font();
-      content: 'd';
-      font-size: 80%;
-      background: black;
-      border-radius: 5px;
-      padding: 0.2em 0.3em;
-      margin-left: 0.2em;
+      display: inline-block;
+      height:14px;
```
thinkh: I would probably make it 1-2 px smaller so that the d has the same height as the title. But we can try that tomorrow at your screen.

src/styles/_datavisyn.scss
```diff
@@ -28,24 +28,33 @@ $dataviyn_font: Monda !default;
...
+      width:12px;
+      background: url('~tdp_core/src/assets/icons/datavisyn/datavisyn_newlogo.png') no-repeat 0 0;
```
thinkh: I think we can just name the file _datavisyn_logo.png_. And maybe we can use an SVG version?

src/styles/_datavisyn.scss
thinkh: Btw. there is a space missing after the `:`

src/styles/_datavisyn.scss
```diff
@@ -28,24 +28,33 @@ $dataviyn_font: Monda !default;
...
       @include datavisyn-font();
-      content: 'd';
-      font-size: 80%;
-      background: black;
-      border-radius: 5px;
-      padding: 0.2em 0.3em;
-      margin-left: 0.2em;
+      display: inline-block;
+      height:14px;
+      width:12px;
```
thinkh: Btw. there is a space missing after the `:`

thinkh:
We should check if logos of other applications are still working and not affected by this change.

oltionchampari:
For my app it doesn't exactly work  because height is too big .
Maybe some more changes needed.

Screenshot_2019-10-14

src/styles/_datavisyn.scss
```diff
@@ -28,24 +28,29 @@ $dataviyn_font: Monda !default;
...
+
   .navbar-brand.caleydo_app {
     background: none;
     padding-left: 0;
+    display: flex;
+    align-items: center;
+    margin-left: 0.8em;
 
     &:before {
       @include datavisyn-font();
```
oltionchampari: You don't need to include the font, if the `content` is empty.

PR #195, opened 2019-08-26
Improve session loading/saving menu

thinkh:
**Changes**

* Call it _Analysis Session Management_
* Add "File open" and "File save" icons to it
* Add the current session name as the top entry
* Change the cloud icon into floppy disk icon
* Un-hide session import and export

grafik


anita-steiner:
is this now for all --> I don't think that we can do this because all (also patentbay ..) are effected on this

thinkh:
Yes, that's correct. Then we need to sepearte that somehow. We could use an extension point which allows customizations. We can discuss solutions in person tomorrow.

PR #196, opened 2019-08-26
Improve add column dropdown menu for LineUp

thinkh:
* Rename _Database Columns ..._ into _Annotation Columns ..._
* Add divider for the last two entries called `Special Columns`

grafik


lehnerchristian:
somehow I like "Database Columns" more, because it's clear where the data comes from

the rest is fine for me

thinkh:
I reverted the name to _Database Columns ..._ again. Hence, this PR only contains the grouping of _Special Columns_.

grafik

@lehnerchristian Please check it again and merge it. Thanks.

PR #198, opened 2019-08-27
Release 5.3.0

anita-steiner:
* change phovea_server to 2.3
* add long_description_content_type to setup.py

Form Builder:
* Additional `step` option for numeric text input fields to control the step size
* Set custom classes for FormRadio elements
* Set custom classes for FormSelect3 elements

rest.ts
* Fixed typings

Security:
* Security for DBViews can be disabled
* Security for views in the frontend can be disabled

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,3 +1,3 @@
 /modules.rst
 /phovea_server.rst
-
+/tdp_core.rst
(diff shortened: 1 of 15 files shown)


anita-steiner:
@lehnerchristian could you please add the additional changes as comment - for the release description

PR #201, opened 2019-08-29
Annotate actionFunction extensions with tdp_matomo

thinkh:
Requires PR https://github.com/datavisyn/tdp_matomo/pull/3 to see the tracking in action. Without this plugin, nothing will happen.

## Changes

* Before this PR the matomo tracking for `actionFunctions` of tdp_core has been hard-coded in [tdp_matomo](https://github.com/datavisyn/tdp_matomo/blob/981f6df5b1ec1f4b81df78eba0f4c945aa12f2cd/src/actions.ts#L12-L27)
* Now the matomo plugin is looking for annotated phovea extensions of extension point `actionFunction` and tracks these functions
* The new properity is used by tdp_matomo
* The properties of the configuration follow the `IMatomoEvent` interface

### DIFF
--- a/src/phovea.ts
+++ b/src/phovea.ts
@@ -6,16 +6,26 @@
 import {IRegistry} from 'phovea_core/src/plugin';
 
 export default function (registry: IRegistry) {
-  function actionFunction(id: string, factory: string, loader: () => any) {
-    registry.push('actionFunction', id, loader, { factory });
+  function actionFunction(id: string, factory: string, loader: () => any, options?: {}) {
+    registry.push('actionFunction', id, loader, { factory, ...options });
   }
 
   function actionCompressor(id: string, factory: string, matches: string, loader: () => any) {
     registry.push('actionCompressor', id, loader, { factory, matches });
   }
 
-  actionFunction('tdpInitSession', 'initSessionImpl', () => System.import('./internal/cmds'));
-  actionFunction('tdpSetParameter', 'setParameterImpl', () => System.import('./internal/cmds'));
+  actionFunction('tdpInitSession', 'initSessionImpl', () => System.import('./internal/cmds'), {
+    tdp_matomo: {
+      category: 'session',
+      action: 'init'
+    }
+  });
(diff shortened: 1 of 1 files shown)


src/phovea.ts
```diff
@@ -6,16 +6,26 @@
...
   }
 
   function actionCompressor(id: string, factory: string, matches: string, loader: () => any) {
     registry.push('actionCompressor', id, loader, { factory, matches });
   }
 
-  actionFunction('tdpInitSession', 'initSessionImpl', () => System.import('./internal/cmds'));
-  actionFunction('tdpSetParameter', 'setParameterImpl', () => System.import('./internal/cmds'));
+  actionFunction('tdpInitSession', 'initSessionImpl', () => System.import('./internal/cmds'), {
+    tdp_matomo: {
```
lehnerchristian: I think the key `tdp_matomo` should be changed, because since we have loose coupling with the extensions, `tdp_core` should not know anything about the possible trackers. If we had a Google Analytics tracker, the key would also be required to be named `tdp_matomo`

thinkh: Good point. I used the term `tdp_matomo` to follow the `config.json` approach and I expect that the configuration properties (i.e., `category`, `action`, `value`, etc.) are specific to Matomo.

If you think that we can re-use the properties for instance for another analytics tool, then it will probably make sense to rename it to something more general like `analytics`.

thinkh: I found out that Google Analytics has basically the same properites as Matomo:

[Matomo](https://matomo.org/docs/event-tracking/)
```js
trackEvent(category, action, [name], [value])
```

[Google Analytics](https://developers.google.com/analytics/devguides/collection/analyticsjs/events)
```js
ga('send', 'event', [eventCategory], [eventAction], [eventLabel], [eventValue], [fieldsObject]);
```

So I'm fine with renaming `tdp_matomo` to `analytics`.

IMO the only question left to discuss is: How would a developer know where to look for a documentation for the `analytics` property? How would she/he get the link to the other plugin? 

(If you want we can also discuss this on Monday in person.)

PR #211, opened 2019-09-11
Update to LineUp v3.2 and adapt to changes

### LINKED ISSUE #210: Adapt to removed `color` property in upcoming LineUp release
The upcoming [LineUp release v3.1.8](https://github.com/lineupjs/lineupjs/tree/release_3.1.8) changes the color assignment of columns. 
PR #102 removed the `color` property from `Column` and setting the the `color` property in the column description is deprecated. Instead the new property `colorMapping` should be used. The color mapping was introduced in PR lineupjs/lineupjs#105.

The `color` property is assigned in https://github.com/datavisyn/tdp_core/blob/2dad446ada8e2591c23b551caf53fa76c8bdb45f/src/lineup/ARankingView.ts#L392

Further the `color` property is accessed in LineUpColor and used to remove colors that are already in use in this ranking.

https://github.com/datavisyn/tdp_core/blob/2dad446ada8e2591c23b551caf53fa76c8bdb45f/src/lineup/internal/LineUpColors.ts#L12-L22

The alternative approach to access the color is `d.getColor(<row>)` which requires a row object that is not available at that time. Hence, we cannot provide this feature and colors might get duplicated.

thinkh:
Fixes #210 

### Summary

* Update to [LineUp v3.2.0](https://github.com/lineupjs/lineupjs/releases/tag/v3.2.0)
* Use `colorMapping` property for column description as default
* Remove check of used colors in ranking, because `Column.color` was removed is only available on a per row basis (and we do not have a row).

### DIFF
--- a/src/lineup/internal/LineUpColors.ts
+++ b/src/lineup/internal/LineUpColors.ts
@@ -7,19 +7,7 @@ export default class LineUpColors {
    * Map that assigns each selection ID a color, which is used as color for columns
    */
   private readonly colorMap = new Map<number, {color: string, items: number}>();
-  private colors: string[];
-
-  init(ranking: Ranking) {
-    const colors = scale.category10().range().concat(scale.category20().range().filter((_d,i) => i % 2 === 1));
-    // remove colors that are already in use from the list
-    ranking.flatColumns.forEach((d) => {
-      const i = colors.indexOf(d.color);
-      if (i > -1) {
-        colors.splice(i, 1);
-      }
-    });
-    this.colors = colors;
-  }
+  private colors: string[] = scale.category10().range().concat(scale.category20().range().filter((_d,i) => i % 2 === 1));
 
   getColumnColor(id: number): string {
     if (id < 0) {
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "@types/d3": "3.5.36",
     "@types/select2": "4.0.41",
     "d3": "3.5.17",
-    "lineupjs": "~3.2.0",
+    "lineupjs": "~3.2.1",
     "phovea_clue": "~2.2.0",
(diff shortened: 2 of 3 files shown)


thinkh:
@lehnerchristian or @dg-datavisyn I release LineUp [v3.2.0](https://github.com/lineupjs/lineupjs/releases/tag/v3.2.0) and [v3.2.1](https://github.com/lineupjs/lineupjs/releases/tag/v3.2.1). Please continue and review and merge this PR.

src/lineup/internal/LineUpColors.ts
```diff
@@ -7,19 +7,7 @@ export default class LineUpColors {
...
   private readonly colorMap = new Map<number, {color: string, items: number}>();
-  private colors: string[];
-
-  init(ranking: Ranking) {
-    const colors = scale.category10().range().concat(scale.category20().range().filter((_d,i) => i % 2 === 1));
-    // remove colors that are already in use from the list
-    ranking.flatColumns.forEach((d) => {
-      const i = colors.indexOf(d.color);
-      if (i > -1) {
-        colors.splice(i, 1);
```
lehnerchristian: don't we need to remove the colors which are already in use anymore?

thinkh: In LineUp v3.2 the `color` property was removed from the `Column`. Only `NumericalColumn` support colors in form of `colorMapping`. The difference here: the color is determined on a per row basis, as each bar can have a different color. When initializing the columns, we don't have rows and _one_ color. Hence, we can remove the check for duplicate colors.

thinkh: Side note: For numerical columns every of the 20 colors are still used once, but it will not be checked what colors already added columns have. 

grafik

PR #215, opened 2019-09-17
removed targetIDValue again

lehnerchristian:
the reason to remove the targetIDValue is simply, because developers are already used to use `this.chooser.chosen().name` to get whatever is selected and not `this.chooser.chosen().targetIDValue`

the name inside the data attribute can now have a different value of the name, which is actually shown to the user (the label)

see also
```
return <IFormSelectOptionGroup>{
        name,
        children: group.map((d, j) => ({
          name: groupNames[j], // <-- this is the label shown to the user
          value: String(d),
          data: {
            id: d,
            name: originalTargetNames[j] // <-- this is the original ID
          }
        }))
```

the label (name attribute outside the `data` attribute) and the original ID (name attribute in the `data` attribute) may be the same of no `readableTargetIDType` is defined or different if the IDs are converted to a `readableTargetIDType`.

I think this makes more sense instead of adding an extra attribute to be used in case a `readableTargetIDType` is defined

I think we should discuss this together

### DIFF
--- a/src/views/SelectionChooser.ts
+++ b/src/views/SelectionChooser.ts
@@ -67,7 +67,7 @@ export default class SelectionChooser {
     return this.updateImpl(selection, true);
   }
 
-  chosen(): { id: number, name: string, targetIDValue: string } | null {
+  chosen(): { id: number, name: string } | null {
     const s = this.accessor(this.formID).value;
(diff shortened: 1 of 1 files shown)


dg-datavisyn:
Tested

PR #219, opened 2019-09-24
pass options from CompositeView to detail views, since they may need …

### LINKED ISSUE #223: CompositeView needs to pass additional options to its components, i.e. detail views
the problem is that detail views can get additional options if they need them (such as an AppState if they need to operate on the state, etc.). also the CompositeView gets the additional options, but it needs to pass them to the detail views, because they need the options in reality, not the CompositeView itself

see also https://github.com/datavisyn/tdp_core/blob/master/src/extensions.ts#L307

lehnerchristian:
…them

closes #223

### DIFF
--- a/src/views/CompositeView.ts
+++ b/src/views/CompositeView.ts
@@ -469,9 +469,10 @@ export default class CompositeView extends EventHandler implements IView {
     const desc = (<ICompositeViewPluginDesc>this.context.desc);
 
     const toEntry = (desc: IElementDesc): Promise<ICompositeInfo> => {
+      const descOptions = desc.options || {};
       return Promise.resolve(desc.loader()).then((instance) => (<ICompositeInfo>{
         key: desc.key,
-        options: desc.options,
+        options: Object.assign({}, descOptions, this.options), // also pass the view options from the ViewWrapper to all views
         create: getFactoryMethod(instance, desc.factory || 'create')
       }));
     };


thinkh:
@lehnerchristian Is there a related issue for this PR?

lehnerchristian:
> @lehnerchristian Is there a related issue for this PR?

nope, this issue came up on the fly yesterday. the problem is that detail views can get additional options if they need them (such as an AppState if they need to operate on the state). also the CompositeView gets the additional options, but it needs to pass them to the detail views, because they need the options in reality, not the CompositeView itselt.

see also https://github.com/datavisyn/tdp_core/blob/master/src/extensions.ts#L307

I'll create a ticket with this text and link it

lehnerchristian:
updated the description

src/views/CompositeView.ts
```diff
@@ -469,9 +469,10 @@ export default class CompositeView extends EventHandler implements IView {
     const desc = (<ICompositeViewPluginDesc>this.context.desc);
 
     const toEntry = (desc: IElementDesc): Promise<ICompositeInfo> => {
+      const descOptions = desc.options || {};
       return Promise.resolve(desc.loader()).then((instance) => (<ICompositeInfo>{
         key: desc.key,
-        options: desc.options,
+        options: Object.assign({}, descOptions, this.options), // also pass the view options from the ViewWrapper to all views
```
dg-datavisyn: Is this downward compatible?

PR #225, opened 2019-09-27
Fix wrong CSS classes for Select3

### LINKED ISSUE #224: Different CSS classes for Select3 break the Ordino tour
* Release number or git hash: c0c01ed35dc10de60a3a5a542a8aa2e09bb4b91c
* Web browser version and OS: Chrome 77
* Environment (local or deployed): both

### Steps to reproduce

1. Start local Ordino with tdp_core develop branch
2. Click help icon (upper right)
3. Start _Ordino Welcome Tour_
4. Click through tour until step 12

-> The two cell lines HCC-827 and BT-20 are supposed to be selected. This is not working.

Verify with stable release:

1. Open https://ordino.caleydoapp.org
2. Click help icon (upper right)
3. Start _Ordino Welcome Tour_
4. Click through tour until step 12

-> Works until the end

### Observed behavior

Comparing the two select fields shows that the classes for the select3 are different.

Ordino public (left) and local version (right)

grafik

### Expected behavior

* The CSS classes should be equal to the last release and the Ordino tour should work (again).

thinkh:
Fixes #224 

### Summary

* Fix `AFormElement.setAttribute()` that CSS classes are added one-by-one (using D3 `classed`) instead of replacing the whole class attribute

**Before this PR**

Ordino public (left) and local version (right)

grafik

**After this PR**

Ordino public (left) and local version (right)

grafik

With this change the Ordino tour is working (again) as expected.

ordino-fixed-tour

### DIFF
--- a/src/form/elements/FormSelect3.ts
+++ b/src/form/elements/FormSelect3.ts
@@ -40,8 +40,8 @@ export default class FormSelect3 extends AFormElement<IFormSelect3> {
    * @param elementDesc The form element description
    * @param pluginDesc The phovea extension point description
    */
-  constructor(form: IForm, desc: IFormSelect3, readonly pluginDesc: IPluginDesc) {
-    super(form, desc, pluginDesc);
+  constructor(form: IForm, elementDesc: IFormSelect3, readonly pluginDesc: IPluginDesc) {
+    super(form, elementDesc, pluginDesc);
 
     this.isMultiple = (pluginDesc.selection === 'multiple');
   }
--- a/src/form/elements/FormSelect3.ts
+++ b/src/form/elements/FormSelect3.ts
@@ -60,9 +60,6 @@ export default class FormSelect3 extends AFormElement<IFormSelect3> {
     const options = Object.assign(this.elementDesc.options, {multiple: this.isMultiple});
     this.select3 = new Select3(options);
     this.$node.node().appendChild(this.select3.node);
-
-    this.elementDesc.attributes.clazz = this.elementDesc.attributes.clazz.replace('form-control', ''); // filter out the form-control class, because the border it creates doesn't contain the whole element due to absolute positioning and it isn't necessary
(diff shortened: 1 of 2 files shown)


src/form/elements/FormSelect3.ts
```diff
@@ -60,9 +60,6 @@ export default class FormSelect3 extends AFormElement<IFormSelect3> {
     const options = Object.assign(this.elementDesc.options, {multiple: this.isMultiple});
     this.select3 = new Select3(options);
     this.$node.node().appendChild(this.select3.node);
-
-    this.elementDesc.attributes.clazz = this.elementDesc.attributes.clazz.replace('form-control', ''); // filter out the form-control class, because the border it creates doesn't contain the whole element due to absolute positioning and it isn't necessary
-    this.setAttributes(this.$node.select('.select3'), this.elementDesc.attributes);
```
thinkh: The lines where already there in the previous release, but seem to work incorrectly. In this (develop) release they are working correctly, but result in wrong element attributes. Hence, I removed these two lines to get the same DOM as in the previous release.  

Alternatively, we could replace `form-control` with `select3` to persist the CSS class.

```
this.elementDesc.attributes.clazz = this.elementDesc.attributes.clazz.replace('form-control', 'select3');
```

@lehnerchristian Please choose which version is correct in your opinion. IMO it's the version I commited.

lehnerchristian: so no custom classes can be applied to FormSelect3? this was done through `this.setAttributes`

some time ago I had some problems with the `form-control` class and Select3 where the border was not as wide as the Select3 element itself. don't you have the same error?

lehnerchristian: my error with the border was probably caused by another CSS rule applying an absolute positioning

but being able to set custom CSS classes would still be nice if it's necessary

thinkh: I changed the code that classes are now added using the D3 `classed` method. Hence existing CSS classes are kept.

PR #231, opened 2019-10-14
Enable AView to process 1:n mappings by id and by name

### LINKED ISSUE #230: Extending AView to able to resolve 1:n mappings

dg-datavisyn:
Closes #230 

### Summary

The resolveSelection method of AView internally relies on the numeric _id value and only maps to the first targetIdType element. This value is not always known to external MappingProviders. So a mapping by name is also needed. Additionally, the AView is now also able to resolve 1:n mappings and not just 1:1 mappings.

This change causes a new minor version.

### DIFF
--- a/src/views/AView.ts
+++ b/src/views/AView.ts
@@ -12,7 +12,7 @@ import {
   EViewMode, ISelection, isSameSelection, IView, IViewContext, VIEW_EVENT_ITEM_SELECT,
   VIEW_EVENT_LOADING_FINISHED, VIEW_EVENT_UPDATE_ENTRY_POINT, VIEW_EVENT_UPDATE_SHARED
 } from './interfaces';
-import {resolveIds, resolveAllNames, resolveAllIds} from './resolve';
+import {resolveIds, resolveAllNames, resolveAllIds, resolveNames} from './resolve';
 import {DEFAULT_SELECTION_NAME} from '../extensions';
 import {IForm} from '../form/interfaces';
 
@@ -271,7 +271,7 @@ export abstract class AView extends EventHandler implements IView {
    * @returns {Promise<string[]>}
    */
   protected resolveSelectionByName(idType = this.idType): Promise<string[]> {
-    return resolveIds(this.selection.idtype, this.selection.range, idType);
+    return resolveNames(this.selection.idtype, this.selection.range, idType);
   }
 
    /**
--- a/src/views/AView.ts
+++ b/src/views/AView.ts
@@ -12,7 +12,7 @@ import {
   EViewMode, ISelection, isSameSelection, IView, IViewContext, VIEW_EVENT_ITEM_SELECT,
   VIEW_EVENT_LOADING_FINISHED, VIEW_EVENT_UPDATE_ENTRY_POINT, VIEW_EVENT_UPDATE_SHARED
 } from './interfaces';
-import {resolveIds} from './resolve';
+import {resolveIds, resolveAllNames, resolveAllIds} from './resolve';
 import {DEFAULT_SELECTION_NAME} from '../extensions';
 import {IForm} from '../form/interfaces';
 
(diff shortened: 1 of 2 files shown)


src/views/AView.ts
```diff
@@ -259,13 +259,39 @@ export abstract class AView extends EventHandler implements IView {
...
+  /**
+   * resolve the name of the current input selection
+   * @returns {Promise<string[]>}
+   */
+  protected resolveSelectionByName(idType = this.idType): Promise<string[]> {
+    return resolveNames(this.selection.idtype, this.selection.range, idType);
+  }
+
+  /**
+   * resolves the ids of the current input selection to all 1:n related names, not just the first one like resolveSelection does
```
thinkh:
```suggestion
   * resolve the ids of the current input selection to all 1:n related names, not just the first one like `resolveSelection` does
```

Either use always `resolve` or always `resolves`.

src/views/AView.ts
```diff
@@ -259,13 +259,39 @@ export abstract class AView extends EventHandler implements IView {
...
+   * @returns {Promise<string[]>}
+   */
+  protected resolveMultipleSelections(idType = this.idType): Promise<string[][]> {
+    return resolveAllIds(this.selection.idtype, this.selection.range, idType);
+  }
+
+  /**
+   * resolves the names of the current input selection to all 1:n related names, not just the first one like resolveSelection does
```
thinkh:
```suggestion
   * resolve the names of the current input selection to all 1:n related names, not just the first one like `resolveSelectionByName` does
```

Same here

src/views/resolve.ts
```diff
@@ -33,3 +33,38 @@ export function resolveIds(fromIDType: IDType, ids: Range | number[], toIDType:
...
+  const target = toIDType === null ? fromIDType : resolve(toIDType);
+  if (fromIDType.id === target.id) {
+    // same just unmap to name
+    return fromIDType.unmap(ids);
+  }
+  // assume mappable
+  return fromIDType.unmap(ids).then((names) => {
+    return fromIDType.mapNameToFirstName(names, target);
+  });
+
```
thinkh: Empty row can be removed.

src/views/resolve.ts
```diff
@@ -33,3 +33,38 @@ export function resolveIds(fromIDType: IDType, ids: Range | number[], toIDType:
   // assume mappable
   return fromIDType.mapToFirstName(ids, target);
 }
+
+export function resolveNames(fromIDType: IDType, ids: Range | number[], toIDType: IDType | string = null): Promise<string[]> {
```
thinkh: Please add some JSDoc the new (and maybe also the existing) functions

src/views/resolve.ts
```diff
@@ -33,3 +33,38 @@ export function resolveIds(fromIDType: IDType, ids: Range | number[], toIDType:
...
+    // same just unmap to name
+    return fromIDType.unmap(ids).then((ids) => [ids]);
+  }
+  // assume mappable
+  return fromIDType.unmap(ids).then((names) => {
+    return fromIDType.mapNameToName(names, target);
+  });
+}
+
+export function resolveAllIds(fromIDType: IDType, ids: Range | number[], toIDType: IDType | string = null): Promise<string[][]> {
```
thinkh: I would move this function below `resolveIds` and before `resolveNames`.

thinkh: What is the difference between this function (`resolveAllIds`) and `resolveIds`?

thinkh: I see it's the `mapToName` vs. `mapToFirstName`. Please add some documentation for each function to clarify what it does/returns. Thanks.

src/views/resolve.ts
dg-datavisyn: Removed

src/views/resolve.ts
dg-datavisyn: Documentation added to all functions

Issue #233, opened 2019-10-24
Create extension point to allow patching of LineUp columns


dvmichaelpuehringer:
Currently, when a score is created, it automatically calculates `min` and `max` for `NumberColumn`, the `categories` for `CategoricalColumn` and so on. It does that by patching the columns after the score values are loaded (see [here](https://github.com/datavisyn/tdp_core/blob/master/src/lineup/internal/column.ts#L97-L137)). Personally, I think the current implementation is **very** controversial, as it requires an `<any>` cast of the columns because LineUp does not support setting this data dynamically. 

Nonetheless, sometimes it is required that custom columns, which are only available in the specific apps, are patched in a way such that they are usable as scores. Therefore, I propose to create an extension point which receives all information supplied to the `patchColumn` function with the ability to update it as required.

PR #234, opened 2019-10-24
Add LineUp column patcher extension

dvmichaelpuehringer:
See #233 for a more information on the problem and why we need to solve it. 

**Usage** 
The column patcher is a simple function, included as extension point with the type `tdpScoreColumnPatcher`. The following example would modify an (already internally patched) `NumberColumn` score. The actual use case is to patch your own column, which is only accessible from within an app and has an known interface (with methods such as `setCategories(...)`). 

1. Add the extension to your `phovea.ts`: 
```
  registry.push('tdpScoreColumnPatcher', 'another-number-column-patcher', () => System.import('./patcher/AnotherNumberColumnPatcher'), {
    factory: 'patchColumn'
  });
```
2. As we have defined the `factory: 'patchColumn'`, we have to create the file `./patcher/AnotherNumberColumnPatcher.ts` with an exported function `patchColumn`. 
```
export function patchColumn(pluginDesc: IPluginDesc, colDesc: IAdditionalColumnDesc, rows: IScoreRow<any>[], col: Column) {
  if (colDesc.type === 'number' && col instanceof NumberColumn) {
    // Calculate something using the rows, and set it on the column...
  }
}
```
3. Now, each time a score is loaded, your patcher will be called (after the internal patching) to allow your updates to take place.

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -15,7 +15,6 @@ export * from './tour/extensions';
 export const EXTENSION_POINT_TDP_SCORE = 'tdpScore';
 export const EXTENSION_POINT_TDP_SCORE_IMPL = 'tdpScoreImpl';
 export const EXTENSION_POINT_TDP_SCORE_LOADER = 'tdpScoreLoader';
-export const EXTENSION_POINT_TDP_SCORE_COLUMN_PATCHER = 'tdpScoreColumnPatcher';
 export const EXTENSION_POINT_TDP_RANKING_BUTTON = 'tdpRankingButton';
 export const EXTENSION_POINT_TDP_VIEW = 'tdpView';
 export const EXTENSION_POINT_TDP_INSTANT_VIEW = 'tdpInstantView';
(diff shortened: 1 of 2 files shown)


src/extensions.ts
```diff
@@ -15,6 +15,7 @@ export * from './tour/extensions';
 export const EXTENSION_POINT_TDP_SCORE = 'tdpScore';
 export const EXTENSION_POINT_TDP_SCORE_IMPL = 'tdpScoreImpl';
 export const EXTENSION_POINT_TDP_SCORE_LOADER = 'tdpScoreLoader';
+export const EXTENSION_POINT_TDP_SCORE_COLUMN_PATCHER = 'tdpScoreColumnPatcher';
```
thinkh: I would use the new extension point id notation `EP_TDP_CORE_XXXX = 'epTdpCoreXXXXXXX'` (see [further down](https://github.com/datavisyn/tdp_core/pull/234/files#diff-887a9ff607524fd9a4e2cddfbec834c6R37)).

```suggestion
export const EP_TDP_CORE_SCORE_COLUMN_PATCHER = 'epTdpCoreScoreColumnPatcher';
```

thinkh: Maybe we should group the parts of extension points so that it becomes more clear which interfaces belong to which constant:

```ts
export const EXTENSION_POINT_TDP_SCORE_COLUMN_PATCHER = 'tdpScoreColumnPatcher';

export interface IScoreColumnPatcherExtension {
  factory(pluginDesc: IPluginDesc, colDesc: IAdditionalColumnDesc, rows: IScoreRow<any>[], col: Column): Promise<void>;
}

export interface IScoreColumnPatcherExtensionDesc extends IPluginDesc {
  load(): Promise<IPlugin & IScoreColumnPatcherExtension>;
}
```

thinkh: Please add a comment what the extension point does. You don't have to add comments about the `@registryParam` and `@factoryParam` as you described it as interface below.

Issue #236, opened 2019-10-28
Downloaded Excel files do not contained extended header


thinkh:
## Environment
 * Browser: Chrome
 * deployed/local: BI Ordino dev

## Steps to reproduce the bug
1. Open list of cell lines
2. Add `Gene Score (Single)`
3. Download data as Excel file
4. Download data as CSV file

-> The CSV file header contains all the information about the added column (gene name, data type, data sub type), however, the Excel file is lacking this information.

It would be nice if this could be added.

We can confirm this. Once it is fixed it should look like this:

![grafik.png](https://images.zenhubusercontent.com/599562848f62dc7798c31c64/01304c1a-95a7-4e71-bbcc-e1128440625b)

The missing information must be added to the _from_json_ request:

![grafik.png](https://images.zenhubusercontent.com/599562848f62dc7798c31c64/7bd55111-7684-4d3f-8a48-936efa634c23)

PR #237, opened 2019-10-29
Add extended header to Excel files

oltionchampari:
closes datavisyn/tdp_core#236

### DIFF
--- a/src/lineup/internal/export.ts
+++ b/src/lineup/internal/export.ts
@@ -7,6 +7,10 @@ function isDateColumn(column: Column) {
   return column.desc.type === 'date';
 }
 
+function getColName(column: Column) {
+  return column.label + (column.description ? '\n' + column.description : '');
+}
+
 export function exportRanking(columns: Column[], rows: IDataRow[], separator: string) {
   //optionally quote not numbers
   const escape = new RegExp(`["]`, 'g');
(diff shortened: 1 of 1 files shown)


src/lineup/internal/export.ts
```diff
@@ -48,14 +48,14 @@ export function exportxlsx(columns: Column[], rows: IDataRow[]) {
   const converted = rows.map((row) => {
     const r: any = {};
     for (const col of columns) {
-      r[col.label] = isNumberColumn(col) ? col.getRawNumber(row) : col.getValue(row);
+      r[`${col.label}${col.description ? `\n${col.description}` : ''}`] = isNumberColumn(col) ? col.getRawNumber(row) : col.getValue(row);
```
thinkh: - [ ] Replace template string with normal concatinated string, since we only need a string for the `\n`
- [ ] Extract the generation of the key/name into a small separate function (similar to the `isDateColumn()` above)
- [ ] Use the new key function in line 29, 51, and 58

oltionchampari: Done !

oltionchampari:
> Is it also necessary to change line 44? (I cannot attach a comment there.)
> 
> ```
> r[col.label] = isNumberColumn(col) ? col.getRawNumber(row) : col.getExportValue(row, 'json');
> ```

Yes it makes it consistent for exportToJson also 

Screenshot from 2019-10-30 08-13-42

Screenshot from 2019-10-30 08-13-22

PR #245, opened 2019-11-08
Fix write access for public sessions

### LINKED ISSUE #176: Session Management: No write access for shared sessions
The following things does not work:
* no difference between read and write access
* give read access to a group: everyone can see it - does not matter which group the second person belongs to

thinkh:
Closes #176 

### Summary 

Independent of the granted permissions users were only allowed to clone a persistent session. Now, the write permission is checked and the select button is displayed. For read permission only the select button is not shown.

### Test

1. Created three sessions:
   * Session shared with zen_austin as buddy
     grafik
   * Public session everyone can write
     grafik
   * Public session everyone can read
     grafik
2. Login as user _zen_austin_
3. Open the persistent session section
  grafik

### DIFF
--- a/src/SessionList.ts
+++ b/src/SessionList.ts
@@ -73,10 +73,10 @@ abstract class ASessionList {
     });
     $enter.select('a[data-action="select"]').on('click', (d) => {
       stopEvent();
-      if (!canWrite(d)) {
-        manager.cloneLocal(d);
-      } else {
+      if (canWrite(d)) {
         manager.loadGraph(d);
+      } else {
+        manager.cloneLocal(d);
       }
       return false;
     });
--- a/src/SessionList.ts
+++ b/src/SessionList.ts
@@ -308,11 +308,19 @@ export class PersistentSessionList extends ASessionList {
         {
           const $tr = $parent.select('#session_others tbody').selectAll('tr').data(otherworkspaces);
 
-          const $trEnter = $tr.enter().append('tr').html(`
-              <td></td>
-              <td></td>
-              <td></td>
-              <td>${ASessionList.createButton('clone')}</td>`);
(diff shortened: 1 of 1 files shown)


thinkh:
@lehnerchristian Thanks for your review. Do you mean the following behavior?

I opened a public session with write access and the _Save Sesssion_ menu item is disabled. 

grafik

Looking at the code the behavior seems to be intended:

https://github.com/datavisyn/tdp_core/blob/f1ba111c3b499a70bcbabe638876e2b6a3cee0f1/src/internal/EditProvenanceGraphMenu.ts#L29-L43

_Save Session_ migrates a temporary session to a persistent session. New states (i.e., nodes and actions) of the persistent session are synchronized automatically. Hence, there is no need for the save action anymore. On the one hand, we could probably hide the save session button, but on the other hand it might confuse the user. In both cases it might be a new issue and PR. 😉

lehnerchristian:
> @lehnerchristian Thanks for your review. Do you mean the following behavior?
> 
> I opened a public session with write access and the _Save Sesssion_ menu item is disabled.
> 
> grafik
> 
> Looking at the code the behavior seems to be intended:
> 
> https://github.com/datavisyn/tdp_core/blob/f1ba111c3b499a70bcbabe638876e2b6a3cee0f1/src/internal/EditProvenanceGraphMenu.ts#L29-L43
> 
> _Save Session_ migrates a temporary session to a persistent session. New states (i.e., nodes and actions) of the persistent session are synchronized automatically. Hence, there is no need for the save action anymore. On the one hand, we could probably hide the save session button, but on the other hand it might confuse the user. In both cases it might be a new issue and PR. wink


ok, didn't know that when the provenance graph changes it is automatically updated in the DB. thanks for the clarification :slightly_smiling_face:

PR #264, opened 2019-12-09
Move IResult, ISearchProvider and ISearchProviderDesc from dTiles to tdp_core

dvvanessastoiber:
Fixes https://github.com/datavisyn/dTiles/issues/218#event-2857990620

### Summary
* create new folder _src/public/search_
* move `IResult`, `ISearchProvider` and `ISearchProviderDesc` to new folder _src/public/search_
* create _index.ts_ for export

### DIFF
--- a/src/public/search/SearchProviderInterfaces.ts
+++ b/src/public/search/SearchProviderInterfaces.ts
@@ -1,0 +1,82 @@
+/*********************************************************
+ * Copyright (c) 2018 datavisyn GmbH, http://datavisyn.io
+ *
+ * This file is property of datavisyn.
+ * Code and any other files associated with this project
+ * may not be copied and/or distributed without permission.
+ *
+ * Proprietary and confidential. No warranty.
+ *
+ *********************************************************/
+
+
+import {IPluginDesc} from 'phovea_core/src/plugin';
+import {IDType} from 'phovea_core/src/idtype';
+
+/**
+ * a search result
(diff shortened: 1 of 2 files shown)


src/public/search/SearchProviderInterfaces.ts
```diff
@@ -0,0 +1,82 @@
+/*********************************************************
+ * Copyright (c) 2018 datavisyn GmbH, http://datavisyn.io
```
thinkh: @lehnerchristian Should we update the year or remove this copyright banner entirely? I checked the files and noticed a lot of inconsistencies. Beside this file there are only 2 more files with this copyright banner.

thinkh:
@dg-datavisyn Do we need more documentation (code documentation or wiki) for these interfaces or is it fine like this?

src/public/search/SearchProviderInterfaces.ts
thinkh: We agreed with @dg-datavisyn that we will remove the banner for now and update all our files with a common banner later.

thinkh:
There is a [wiki documentation](https://wiki.datavisyn.io/tdp/tutorials/how-to-add-a-target-360-search-provider) available.

thinkh:
@lehnerchristian is it ok for you if we merge this PR? Since we are just adding something it should not break anything.

PR #267, opened 2019-12-12
Added lazy loaded column interface to functions like addLazyColu…

dvmichaelpuehringer:
Moved the inline-return-value-definition of `addLazyColumn` to a separate interface which is now used as return-value for functions such as `addTrackedScoreColumn` to improve type safety.

### DIFF
--- a/src/base/AEmbeddedRanking.ts
+++ b/src/base/AEmbeddedRanking.ts
@@ -9,6 +9,7 @@ import {EXTENSION_POINT_TDP_SCORE_IMPL} from '../extensions';
 import {get as getPlugin} from 'phovea_core/src/plugin';
 import {IServerColumnDesc} from '../rest';
 import {IFormElementDesc} from '../form';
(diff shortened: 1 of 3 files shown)


thinkh:
@lehnerchristian What the status here?

thinkh:
@lehnerchristian Would you tag this PR as **minor**, because a new TS interface was added, or as **patch**, because nothing has changed to the outside?

dvmichaelpuehringer:
@thinkh the interface is just extracted from the existing return value for reusability, it does not really change anything (internally/externally) as it was inferable by TS before as well.

lehnerchristian:
I've tested it yesterday in two apps and it worked for me --> at least I had no compiler errors

However I don't have some apps installed, so I cannot test them there

lehnerchristian:
I'd say it's a patch, isn't it?

thinkh:
Yes, let's label it as patch. Anyway there will be a major release next time. 😉

PR #272, opened 2019-12-13
check if arguments are present

lehnerchristian:
this basically undoes the latest changes in the following line: https://github.com/datavisyn/tdp_core/blob/master/tdp_core/db.py#L271

@anita-steiner , I'm currently not sure why these changes were for. undoing it worked in my tests. if we need to `not info or not info.list_as_tuple` check I can undo my changes and add an else branch to this conditional. however I think just aborting the request as we have no given parameters for the required argument makes sense


lehnerchristian:
in an f2f code review we agreed on keeping the old version of the `elif` and provide a fallback value

PR #276, opened 2019-12-19
Use datavisyn logo as loading icon

lehnerchristian:
This PR changes the loading icon to the new datavisyn loading icon. 
This is done by overriding SCSS variables (where possible), and overriding `.phovea-busy` (because overriding `$phovea-loading-icon` does not work for some reason?). 

Because we now utilize overriding variables, we should import the datavisyn branding before the actual tdp style. 
This has to be done in the apps individually, often in the `src/style.scss`: 
```
@import "~tdp_core/src/styles/datavisyn";  // Branding before other styles
@import "~tdp_core/src/styles/tdp";
```


thinkh:
Issue https://github.com/Caleydo/ordino/issues/240 describes the test if the Caleydo logo is still shown after this change in Ordino. Result: The Caleydo logo is still shown, so from my side this PR is fine.

anita-steiner:
there was also a problem with the loading order; so please don't merge the pr til we fix it

puehringer:
@thinkh Ordino itself does not import the datavisyn branding at all (https://github.com/Caleydo/ordino/blob/master/src/style.scss), such that the proper scss-overrides should work properly (as they are never loaded, and if they are loaded, it is after the `tdp` style). The only problem is when a dependency imports this style which then _hard_ overrides `.phovea-busy`. 
I would prefer a solution with overriding `$phovea-loading-icon`, maybe you have an idea on how we can accomplish that?

puehringer:
@thinkh After the discussion today, could you link the visyn2020 issue here after it is written?

thinkh:
Here you go: https://github.com/datavisyn/visyn2020/issues/42

anita-steiner:
will be done in a new pr

Issue #257, opened 2019-11-25
Extract Touring/TourDino from tdp_core to a separate repo


thinkh:
The goal of this issue should be the integration of Touring/TourDino into tdp_core.

TourDino adds an additionall button to the `ARankingView` which opens a new panel. The user can statistically evaluate the selections/groups of rows and columns. See also the [video and publication](http://caleydo.org/publications/2019_eurova_tourdino/). 

![grafik](https://user-images.githubusercontent.com/5851088/69532315-d194c680-0f75-11ea-8c8c-71442ed07d5d.png)

Demo: https://tourdino.caleydoapp.org/

The current implementation consists of two parts:

1. [Klaus' touring branch @ tdp_core](https://github.com/datavisyn/tdp_core/compare/develop...keckelt/122/touring?expand=1)
2. [tourdino repo](https://github.com/Caleydo/tourdino)

**Tasks**

- [ ] Create a new `touring` branch
- [ ] Copy directory and files from Klaus' branch to the new touring branch
- [ ] Rename directories to lower case (e.g., `src/lineup/internal/Touring/Tasks` -> `src/lineup/internal/touring/tasks`)
- [x] Try to revert to `es5` as compiler target in _tsconfig.json_ (see https://github.com/datavisyn/tdp_core/compare/develop...keckelt/122/touring?expand=1#diff-e5e546dd2eb0351f813d63d1b39dbc48R4)
- [x] Related to ES5: Enable uglifyjs in Webpack config again (see https://github.com/datavisyn/tdp_core/compare/develop...keckelt/122/touring?expand=1#diff-11e9f7f953edc64ba14b0cc350ae7b9dL354-L424)
- [ ] [TourDino code](https://github.com/Caleydo/tourdino) is integrated in tdp_core 
   * i.e., no dependency to the tourdino repository in _package.json_
   * should be placed under `tdp_core/src/lineup/internal/touring/` 
- [ ] Move classes `ATouringTask`, `RowComparison`, and `ColumnComparison` to separate files
- [ ] Separate and move Cypress dependencies, config and tests (for Ordino) to [Caleydo/ordino](https://github.com/Caleydo/ordino)
- [ ] The Touring button in ARankingView should be configurable (i.e., a show/hide via config flag)
[...]

PR #285, opened 2020-01-09
Create extension point to connect the touring panel/Refactor LineupSidePanel

oltionchampari:
Closes datavisyn/tdp_core#257

Developer documentation:
https://wiki.datavisyn.io/tdp/tutorials/lineup-panel/lineup-side-panel
https://wiki.datavisyn.io/tdp/tutorials/lineup-panel/how-to-add-line-up-panel-tab
https://wiki.datavisyn.io/tdp/tutorials/lineup-panel/how-to-add-a-new-tdp-ranking-button

### Summary
- Refactored and extracted the LineupSidePanel class into several different files to make it more scalable 
- Added tabbing functionality to the side panel in able to have different content inside of it 
- Created extension point to add further tabs to the side panel

paneltab

**To register to the above extension point:**

```typescript
import { EP_TDP_CORE_LINEUP_PANEL_TAB, IPanelTabExtensionDesc } from 'tdp_core/src/extensions';

  registry.push(EP_TDP_CORE_LINEUP_PANEL_TAB, 'statisticalAnalysisPanel', function () {
    return System.import('./TouringPanel');
  }, <IPanelTabExtensionDesc>{
    cssClass: 'fa-calculator',
    title: 'Statistical Analysis',
    order: 10,
    width: '45em',
    shortcut: true
  });
```
Factory method name /entry point can be defined:
```typescript
factory:'<my factory method>'
```
If not defined the default factory method name is 
```typescript
create() 
```
### TODO

- [x] Test side panel in application where it is used in top mode


src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -58,39 +65,46 @@ export default class LineUpPanelActions extends EventHandler {
...
     }
-    this.node.classList.add('tdp-view-lineup');
-    this.collapse = options.enableSidePanel === 'top' || options.enableSidePanel === 'collapsed';
 
     this.init();
+    this.collapse = options.enableSidePanel === 'top' || options.enableSidePanel === 'collapsed';
+  }
+
+  removeTabButtonHighlighting() {
+    this.tabContainer.showDefault();
```
thinkh: It's a bit confusing that the header has to remove the highlight. Especially, because it is directly accessing the CSS classes of the tab. We might have to change the architecture of the tabs a bit. Maybe it helps if the `PanelNavButton` has a reference to his own tab or the `PanelTab` has an (optional) reference to the button that can activate it.

src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -114,14 +128,20 @@ export default class LineUpPanelActions extends EventHandler {
 
   set collapse(value: boolean) {
     this.node.classList.toggle('collapsed', value);
+    if (value && this.options.enableSidePanel !== 'top') {
+      this.removeTabButtonHighlighting();
+    }
```
thinkh: Why is this necessary for a side panel that is not top? Please add some comment to the code.

src/styles/_view_lineup.scss
```diff
@@ -105,18 +105,19 @@ $lu_assets: '~lineupjs/src/assets';
...
   .lu-hierarchy .lu-search::before {
     height: 19px;
   }
 
-  .tdp-view-lineup.lu-side-panel {
-    flex: 0 0 auto;
-    position: relative;
-
+  //////////////////////
+  /////////////////////
```
thinkh: Can this be removed?

src/lineup/internal/utils.ts
```diff
@@ -7,12 +7,18 @@ import {IDataRow} from 'lineupjs';
 import {convertRow2MultiMap, IFormMultiMap, IFormRow} from '../../form';
 import {encodeParams} from 'phovea_core/src/ajax';
 
+
+
+export interface IAccessorFunc<T> {
+  (row: IDataRow): T;
+}
```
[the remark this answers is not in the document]
thinkh: I think this one must be kept and the one in tourdino can import this one (from tdp_core). Please add some source code documentation here.

oltionchampari:
Tested the side panel **top mode**  in patentbay  and it works and looks almost the same.
The only difference is that in patentbay a `margin-left:auto` is added which pushes the side panel to the right and with this branch it doesn't work anymore.
To fix it the css selector in patentbay needs to be adapted.
**Original:**
patent1


**Current branch:**
patent2

src/lineup/internal/panel/PanelButton.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.node.title = title;
+    this.node.addEventListener('click', (evt) => {
+      evt.stopPropagation();
+      evt.preventDefault();
+      this.highlight();
+      onClick();
+    });
+  }
+  private highlight() {
+    if (!this.tabNode.classList.contains('active')) {
```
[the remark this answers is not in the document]
oltionchampari: Refactored implementation.

src/lineup/internal/panel/PanelButton.ts
```diff
@@ -0,0 +1,59 @@
...
+      evt.preventDefault();
+      onClick();
+    });
+  }
+}
+
+export class PanelNavButton implements IPanelButton {
+  readonly node: HTMLElement;
+
+  constructor(parent: HTMLElement, private readonly tabNode: HTMLElement, title: string, linkClass: string, onClick: () => void) {
```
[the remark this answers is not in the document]
oltionchampari: Added documentation.

oltionchampari:
After a short discussion with Dominic, Christian and Michael  about the UI and the behavior of the nav tabs they suggested:

1. The **collapse/uncollapse** button should open the lineup config the first time it gets clicked  
 and the last open tab every other time.

2. There should be an icon for the default **lineup config tab** only in uncollapsed state though.

3. The **nav-tabs** should be on their own line after the default side-panel buttons.
 
Result : 

panel

src/lineup/internal/panel/PanelDownloadButton.ts
```diff
@@ -0,0 +1,55 @@
+import {LocalDataProvider} from 'lineupjs';
+import {exportLogic} from '../export';
+import {IPanelButton} from './PanelButton';
+import i18n from 'phovea_core/src/i18n';
+export default class PanelDownloadButton implements IPanelButton {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

src/lineup/internal/panel/PanelAddColumnButton.ts
```diff
@@ -0,0 +1,31 @@
+import {SearchBox} from 'lineupjs';
+import {ISearchOption} from '../LineUpPanelActions';
+import {IPanelButton} from './PanelButton';
+import i18n from 'phovea_core/src/i18n';
+export default class PanelAddColumnButton implements IPanelButton {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

src/lineup/internal/panel/PanelAddColumnButton.ts
```diff
@@ -0,0 +1,31 @@
...
+  readonly node: HTMLElement;
+
+  constructor(parent: HTMLElement, private readonly search: SearchBox<ISearchOption>) {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

src/lineup/internal/panel/PanelDownloadButton.ts
```diff
@@ -0,0 +1,55 @@
...
+  readonly node: HTMLElement;
+
+  constructor(parent: HTMLElement, private provider: LocalDataProvider, isTopMode: boolean) {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

src/lineup/internal/panel/PanelDownloadButton.ts
```diff
@@ -0,0 +1,55 @@
...
+        const type = link.dataset.t;
+        const onlySelected = link.dataset.s === 's';
+        exportLogic(<any>type, onlySelected, this.provider).then(({content, mimeType, name}) => {
+          this.downloadFile(content, mimeType, name);
+        });
+      };
+    }
+  }
+
+  private downloadFile(content: BufferSource | Blob | string, mimeType: string, name: string) {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

src/lineup/internal/panel/PanelHeader.ts
```diff
@@ -0,0 +1,32 @@
...
+  addButton(button: IPanelButton) {
+    this.buttons = [...this.buttons, button];
+    this.node.appendChild(button.node);
+  }
+
+  removeHighlighting() {
+    const highlightedButton = this.buttons.find((button) => button.node.classList.contains('active'));
+    if (highlightedButton) {
+      highlightedButton.node.classList.remove('active');
+    }
```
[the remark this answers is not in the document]
oltionchampari: Used bootstrap class to add the active styles `PanelNavButton`. We don't need to manually add the active class to the `PanelNavButton`.

src/lineup/internal/panel/PanelRankingButton.ts
```diff
@@ -0,0 +1,20 @@
+import {IPanelButton} from './PanelButton';
+import {LocalDataProvider, Ranking} from 'lineupjs';
+
+export default class PanelRankingButton implements IPanelButton {
```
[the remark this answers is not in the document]
oltionchampari: Commented code.

oltionchampari:
The current status looks like this:

panelTab

On a last refactoring round, changed the component structure to avoid confusing code.

src/lineup/internal/panel/PanelButton.ts
```diff
@@ -33,27 +34,41 @@ export default class PanelButton implements IPanelButton {
...
-    if (!this.tabNode.classList.contains('active')) {
-      this.node.classList.add('active');
 
-    } else {
-      this.node.classList.remove('active');
+  /**
+   * When you click the shortcut button in collapsed mode focus on the navButton
+   */
+  setActive() {
+    const navButtons = Array.from(this.node.parentElement.children);
```
thinkh: Avoid mutating the parent element!

src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -88,25 +88,31 @@ export default class LineUpPanelActions extends EventHandler {
...
+      const sidePanel = new SidePanelTab(this.node, this.searchBoxProvider.createSearchBox(), ctx, doc, options);
       this.panel = sidePanel.panel;
 
+      const listener = () => {
+        this.tabContainer.showTab(sidePanel);
+      };
       this.tabContainer = new PanelTabContainer(this.node);
-      this.tabContainer.addTab(sidePanel);
+      const lineupNavButton = new PanelNavButton(this.tabContainer.node, listener, options, true);
+      this.tabContainer.addTab(sidePanel, lineupNavButton);
```
thinkh: Do we need to create the PanelNavButton outside or could we create the button inside of `addTab`? Because the `PanelTab` (or here `SidePanelTab`) contains already the options.

src/lineup/internal/panel/PanelTabContainer.ts
```diff
@@ -1,64 +1,124 @@
...
   }
 
-  addTab(tab: PanelTab) {
-    this.tabs = [...this.tabs, tab];
-    this.node.appendChild(tab.node);
+  /**
+   * Add nav-tab to the nav-tabs in the correct order.
+   * @param button PanelNavButton instance to add
+   */
+  addNav(nav: PanelNavButton) {
```
thinkh:
```suggestion
  addNav(nav: PanelNavButton) {
```

I would rename `addNav` to `addButton` as we are adding a button to the header.

src/lineup/internal/panel/PanelTabContainer.ts
```diff
@@ -1,64 +1,124 @@
...
+    const sameOrderNav = this.navButtons.find((n) => n.order === nav.order); // find nav with same order
+    this.navButtons = [...this.navButtons, nav].sort((nav1, nav2) => nav1.order - nav2.order);
+    this.node.appendChild(nav.node);
+    const nextSibling = sameOrderNav || this.getNextSibling(this.navButtons, nav); // if same order property append new nav before it
+    if (nextSibling) {
+      this.node.insertBefore(nav.node, nextSibling.node);
+    } else {
+      this.node.appendChild(nav.node);
+    }
```
thinkh: We can shift the sorting to the extension point, as we only want to sort the panel tabs that are added by the extension point. That would simplify the code here.

src/lineup/internal/panel/PanelDownloadButtonContainer.ts
```diff
@@ -2,10 +2,15 @@ import {LocalDataProvider} from 'lineupjs';
 import {exportLogic} from '../export';
 import {IPanelButton} from './PanelButton';
 import i18n from 'phovea_core/src/i18n';
-export default class PanelDownloadButton implements IPanelButton {
+
+
+/**
+ * A button dropdown to download selected/all rows of the ranking
+ */
+export default class PanelDownloadButtonContainer implements IPanelButton {
```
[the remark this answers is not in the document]
thinkh: Please revert the renaming.

oltionchampari:
I retested the side panel in _patentbay_ in top mode again and it works as expected.

src/lineup/internal/panel/PanelDownloadButton.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * A button dropdown to download selected/all rows of the ranking
+ */
+export default class PanelDownloadButton implements IPanelButton {
+  readonly node: HTMLElement;
+
+  constructor(parent: HTMLElement, private provider: LocalDataProvider, isTopMode:boolean) {
+    this.node = parent.ownerDocument.createElement('div');
+    this.node.classList.add('btn-group', 'download-data-dropdown');
+    this.node.innerHTML = `
```
lehnerchristian: why aren't you inserting the elements via `someHTMLNode.insertAdjacentHTML('afterbegin', '...')` [like you're doing here](https://github.com/datavisyn/tdp_core/blob/ochampari/10_tourdino-lineup-panel-button/src/lineup/internal/panel/PanelButton.ts#L54) or vice versa?

src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -217,78 +220,46 @@ export default class LineUpPanelActions extends EventHandler {
...
+          if (this.collapse) {
+            this.collapse = false; // expand side panel
+          }
+          this.tabContainer.showTab(tab);
 
-  protected saveRankingDialog(order: number[]) {
-    editDialog(null, (name, description, sec) => {
-      this.fire(LineUpPanelActions.EVENT_SAVE_NAMED_SET, order, name, description, sec);
+        } else {
+          plugin.load().then((p) => {
```
lehnerchristian: The interface for `factory` is `factory(parent: HTMLElement, provider: LocalDataProvider, desc: IRankingButtonExtensionDesc, idType: IDType, extraArgs: object): Promise<IScoreParam>;` according to `IPanelTabExtension`

so when you add the type `IPanelTabExtension` to p then the method is incompatible to the interface's definition. for example `tab.events` is passed as IDType

src/extensions.ts
```diff
@@ -136,6 +146,40 @@ export interface IScoreColumnPatcherExtensionDesc extends IPluginDesc {
   load(): Promise<IPlugin & IScoreColumnPatcherExtension>;
 }
 
+export interface IPanelTabExtension {
+  desc: IPanelTabExtensionDesc;
+  factory(parent: HTMLElement, provider: LocalDataProvider, desc: IRankingButtonExtensionDesc, idType: IDType, extraArgs: object): Promise<IScoreParam>;
```
lehnerchristian: should the `extraArgs` be optional?

src/lineup/internal/panel/PanelTab.ts
```diff
@@ -0,0 +1,134 @@
...
+
+}
+
+/**
+ * The PanelTab creates a tab component that with can be toggled through the PanelNavButton
+ */
+export class PanelTab {
+
+  readonly node: HTMLElement;
+  readonly events: PanelTabEvents;
```
lehnerchristian: usually we don't have instance variables like this and instead the class itself (`PanelTab` in this case) inherits from `EventHandler`). at least I haven't seen a usage like this yet

@thinkh do you have usages like this in Ordino?

src/lineup/internal/panel/PanelTab.ts
```diff
@@ -0,0 +1,134 @@
...
+    return this.navButton;
+  }
+
+  getShortcutButton(): PanelButton {
+    const onClick = () => {
+      this.navButton.click();
+    };
+
+    // Note: `document.body` is used only for `parent.ownerDocument.createElement()` inside the button
+    return new PanelButton(document.body, this.options.title, 'fa ' + this.options.cssClass + ' shortcut-nav', onClick);
```
lehnerchristian: so `document.body.ownerDocument` === `document`. I see that this parent HTMLElement passing is the pattern but that also does not make much sense I think :)

why not pass a reference to the real parent?

src/lineup/internal/panel/PanelTab.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  readonly panel: SidePanel | null;
+
+  /**
+   * @param parent The parent HTML DOM element
+   * @param search LineUp SearchBox
+   * @param ctx LineUp context
+   * @param doc Document
+   */
+  constructor(parent: HTMLElement, private readonly search: SearchBox<ISearchOption>, ctx: any, doc = document, options?: IPanelTabDesc) {
```
lehnerchristian: the type of `ctx` is not any when a take a look at the `SidePanel` class

src/lineup/internal/panel/PanelTabContainer.ts
```diff
@@ -0,0 +1,114 @@
...
+
+
+
+/**
+ * The header of the PanelTab
+ * Contains the PanelNavButtons that toggle the PanelTab
+ */
+class PanelTabHeader {
+  public node: HTMLElement;
+  public navButtons: PanelNavButton[] = [];
```
lehnerchristian: the PanelNavButtons are never pushed to this array, so it'll always be emtpy. if you push the buttons to this array, please make it private and readonly

src/lineup/internal/panel/SearchBoxProvider.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  get length(): number {
+    return this.searchBoxes.length;
+  }
+
+  /**
+   * Create a new LineUp SearchBox. The instance is added to the internal list and returned.
+   * @returns A new LineUp SearchBox instance
+   */
+  createSearchBox(options?: Partial<ISearchBoxOptions<ISearchOption>>): SearchBox<ISearchOption> {
```
lehnerchristian: I just tested what happens when you pass `undefined` as a parameter to `Object.assign()`. it works, however I think it would be better to have the `options` parameter initialized as an empty object

lehnerchristian:
I have just some minor questions and some comments on typings / unused variables

src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -114,14 +122,27 @@ export default class LineUpPanelActions extends EventHandler {
 
   set collapse(value: boolean) {
     this.node.classList.toggle('collapsed', value);
+
+    // When this.options.enableSidePanel === 'top' `this.collapsed=true` gets called. Attempting to open the PanelTab produces an error.
```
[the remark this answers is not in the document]
oltionchampari: When you have the `enableSidePanel='top'`  `this.collapse `gets set one time.
The problem is that i want the PanelTab to react accordingly when the SidePanel gets collapsed, apart from  when `enableSidePanel='top'` because there is no PanelTabs in top mode.
Running ` this.tabContainer.hideCurrentTab();` throws an error because` this.tabContainer` doesn't exist.

src/lineup/internal/panel/PanelDownloadButton.ts
oltionchampari: No special reason behind it. This can be unified.

src/lineup/internal/LineUpPanelActions.ts
oltionchampari: Thank you, fixing in next commit.

src/extensions.ts
oltionchampari: It should.

src/lineup/internal/panel/PanelDownloadButton.ts
oltionchampari: No special reason, it can be unified.

src/lineup/internal/LineUpPanelActions.ts
oltionchampari: If you could take a look again at the commit to confirm it looks better now.

src/lineup/internal/panel/PanelTab.ts
thinkh: Good point. I changed it.

src/lineup/internal/panel/PanelTab.ts
thinkh: Valid point. 😉 I changed it.

src/lineup/internal/panel/PanelTabContainer.ts
thinkh: Thanks. We forgot to remove this in the last refactoring. I removed the variable now.

src/lineup/internal/panel/SearchBoxProvider.ts
thinkh: Good point! I changed it.

src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -114,14 +123,27 @@ export default class LineUpPanelActions extends EventHandler {
...
+    if (value && this.options.enableSidePanel !== 'top') {
+      this.tabContainer.hideCurrentTab(); // Hide the active PanelTab --> Inform its content to stop updating
+    } else if (this.options.enableSidePanel !== 'top') {
+      this.tabContainer.showCurrentTab(); // Show the last active PanelTab --> Inform its content to start updating again
+    }
```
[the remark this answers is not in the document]
thinkh: I just tried my solution and it still throws an error in the top mode. I suggest to add a `NullTabContainer` which exposes the public functions, but implements no operation. That way we would avoid having lots of if-else scattered in the code.

src/lineup/internal/LineUpPanelActions.ts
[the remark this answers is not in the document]
thinkh: I changed the implementation and added a NullTabContainer, which has no functionality, for the top mode. This avoids these ugly if-else conditions.

src/lineup/internal/panel/PanelTab.ts
thinkh: You are right. Our intention was to pass an event object, so that the extension point can react to some events, and this way avoid to pass the whole tab to the extension point. Otherwise the extension point will have access to functionality it does not need or should not

thinkh: We changed the implementation according to your suggestion and integrated the events into the panel tab.

thinkh:
@lehnerchristian Thanks for your valuable review. We addressed your comments with the last few commits. As we agreed eariler today in a call, I'm gonna merge the this PR even though you haven't revised it again.

PR #288, opened 2020-01-13
Fixed translation strings of internal date utils

puehringer:


thinkh:
Obsolete with https://github.com/datavisyn/tdp_core/pull/298.

PR #295, opened 2020-01-28
Configurable query delay for Select3

lehnerchristian:
The delay when typing into a Select3 box is now configurable, i.e. the delay after the last character has been typed and before the search method of Select2 is called

do we want other Select2 options to be overridden?

for the AJAX options:
- I thought the `cache` option would be good to be configurable, however Select3 builds its own cache
- as far as I can see the URL is never needed, because the the search is performed by the `search` option, which is a function performing the search
- dataType: always JSON?
- transport: Select3's search function

### DIFF
--- a/src/form/elements/Select3.ts
+++ b/src/form/elements/Select3.ts
@@ -147,6 +147,12 @@ export interface ISelect3Options<T extends Readonly<IdTextPair>> {
    * @default ' '
    */
   defaultTokenSeparator: string;
+
+  /**
+   * query delay
+   * @default 250
+   */
+  queryDelay: number;
 }
 
 /**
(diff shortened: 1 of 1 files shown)


src/form/elements/Select3.ts
```diff
@@ -147,6 +147,12 @@ export interface ISelect3Options<T extends Readonly<IdTextPair>> {
    * @default ' '
    */
   defaultTokenSeparator: string;
+
+  /**
+   * query delay
```
thinkh: Repeating the name of the variable in the comment is not really helpful. Could you please describe what this setting does and/or what the effect of a low/high value is?

```suggestion
   * Configures the time span how long to wait after a user has stopped typing before sending the AJAX request.
```

PR #300, opened 2020-02-04
Fix select3 dropdown bug

dvdanielamoitzi:
When opening the project in Mozilla Firefox and adding multiple new devaLines to a project (so that the drop-down is going upwards), the search input did not work anymore (it was not possible to type the query in the input) => see attached files. This is a common problem with bootstrap modals and the select2 library: https://select2.org/troubleshooting/common-problems . Adding the attribute "dropdownParent" fixed it.
Additionally, if the devaLine session is not yet set,  the placeholder of the input field is set to the default placeholder (="Search...").

Before:
buggy

After:
fixed


lehnerchristian:
the dropdown works now, however we have this little positioning bug where the dropdown is placed somewhere above the actual Select2 input element

Screenshot_2020-06-15 Triage

anita-steiner:
> the dropdown works now, however we have this little positioning bug where the dropdown is placed somewhere above the actual Select2 input element (this happens in Chrome as well as in Firefox)
> 
> Screenshot_2020-06-15 Triage

@dvdanielamoitzi could you have a look on it; do we need this pr for the next release?

PR #301, opened 2020-02-05
Parse regular expression filter before storing to provenance graph

oltionchampari:
Closes Caleydo/tdp_bi_bioinfodb#1054

### Summary 
* The cause of the issue was applying` JSON.stringify()` to a RegExp object returns always  `{}`
```javascript
JSON.stringify({regexp: /^B/gm});
// result: "{\"regexp\":{}}"
```
which resulted in the provenance graph not properly storing the  filter value.

* Added a between step that:
  * transforms the regex filter to a string before storing it to the provenance graph 
  * transforms it back to a regular expression when  the replay happens.

### DIFF
--- a/src/lineup/internal/cmds.ts
+++ b/src/lineup/internal/cmds.ts
@@ -323,7 +323,7 @@ function recordPropertyChange(source: Column | Ranking, provider: LocalDataProvi
       return;
     }
 
-    const newSerializedValue = stringifyRegExp(newValue); // serialize possible RegExp object to be properly stored as provenance graph
+    const newSerializedValue = serializeRegExp(newValue); // serialize possible RegExp object to be properly stored as provenance graph
 
     if (source instanceof Column) {
       // assert ALineUpView and update the stats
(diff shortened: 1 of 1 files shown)


thinkh:
@oltionchampari I improved some function and variable names and added more comments.

In my tests it works when I applying a regexp filter in the string column, but it does not work when I add another filter in the categorical column.

ordino-regexp-filter

Please fix this behavior and test it more deeply in combination with other filters and jumping back and forth in the provenance graph.

oltionchampari:
> See previous comment.

Thanks for the feedback and the further optimizations you made!

The problem above was that i wasn't checking for null values.
After extensive testing everything seems to be working as expected.

oltionchampari:
> @oltionchampari I still get an error when I play around with different filters.
> 
> ordino-regexp-filter
> 
> Please test the following provenance graph:
> 
> [Temporary Session 266.json.txt](https://github.com/datavisyn/tdp_core/files/4186585/Temporary.Session.266.json.txt)

@thinkh Cant reproduce it. Maybe you don't have the latest branch .

test

thinkh:
Yes, indeed it works!

thinkh:
@oltionchampari Could please try one more time, if the provenance graph with a `value: {}` in the action will not break anymore.

oltionchampari:
> @oltionchampari Could please try one more time, if the provenance graph with a `value: {}` in the action will not break anymore.

The current solution also works for filter  `value: {}`. The filter `value: {}` is set and the rest of the prov graph steps continue running.

PR #305, opened 2020-02-12
Patch inconsistency of naming renderer getter and setter

oltionchampari:
Closes Caleydo/tdp_bi_bioinfodb#1058

### Summary 
The issue was that : 
- the renderer gets tracked by providing a string from which the name of the event , the getter and the setter are created.
i.e ` recordPropertyChange(col, provider, lineup, graph, 'rendererType');`,
which produces `rendererTypeChanged, getRendererType, setRendererType`,
which of course doesn't work
- in lineup the corresponding names are:
```javascript
rendererTypeChanged
getRenderer()//instead of getRendererType()
setRenderer()//instead of setRendererType()

```
#### To discuss : 
The current implementation in tdp_core expects that the names of the event, setter and getter 
to be similar which is error prone as proved by this issue.

### DIFF
--- a/src/lineup/internal/cmds.ts
+++ b/src/lineup/internal/cmds.ts
@@ -184,7 +184,7 @@ export function setGroupCriteria(provider: IObjectRef<any>, rid: number, columns
 export async function setColumnImpl(inputs: IObjectRef<any>[], parameter: any) {
   const p: LocalDataProvider = await resolveImmediately((await inputs[0].v).data);
   const ranking = p.getRankings()[parameter.rid];
-  let prop = parameter.prop[0].toUpperCase() + parameter.prop.slice(1);
+  const prop = parameter.prop[0].toUpperCase() + parameter.prop.slice(1);
 
   let bak = null;
   const waitForSorted = dirtyRankingWaiter(ranking);
@@ -209,6 +209,7 @@ export async function setColumnImpl(inputs: IObjectRef<any>[], parameter: any) {
         break;
     }
   }
+
   return waitForSorted({
     inverse: setColumn(inputs[0], parameter.rid, parameter.path, parameter.prop, bak)
   });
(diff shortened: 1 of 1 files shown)


src/lineup/internal/cmds.ts
```diff
@@ -184,15 +184,18 @@ export function setGroupCriteria(provider: IObjectRef<any>, rid: number, columns
...
+  }
 
   let bak = null;
   const waitForSorted = dirtyRankingWaiter(ranking);
   let source: Column | Ranking = ranking;
   if (parameter.path) {
     source = ranking.findByPath(parameter.path);
   }
-  ignoreNext = `${parameter.prop}Changed`;
+  // ignoreNext = `${parameter.prop}Changed`;
```
thinkh: Why does this line needs to be commented out?

src/lineup/internal/cmds.ts
```diff
@@ -184,15 +184,18 @@ export function setGroupCriteria(provider: IObjectRef<any>, rid: number, columns
 export async function setColumnImpl(inputs: IObjectRef<any>[], parameter: any) {
   const p: LocalDataProvider = await resolveImmediately((await inputs[0].v).data);
   const ranking = p.getRankings()[parameter.rid];
-  const prop = parameter.prop[0].toUpperCase() + parameter.prop.slice(1);
+  let prop = parameter.prop[0].toUpperCase() + parameter.prop.slice(1);
+  if (prop === 'RendererType') {
+    prop = 'Renderer';
+  }
```
thinkh: I saw that we fixed this issue already in the LineUp v4 branch, but did not backport it here. We should try to make the implementation equal to avoid merge conflicts later.

https://github.com/datavisyn/tdp_core/blob/6581047b02e4c5ac1c8f968e64641c0f93716a27/src/lineup/internal/cmds.ts#L215-L227

Please update the implementation in the lineup-v4 branch accordingly.

src/lineup/internal/cmds.ts
oltionchampari: It is not used anywhere, it is just declared.

src/lineup/internal/cmds.ts
oltionchampari: I adapted the implementation to match that of the lineup4 branch. Further tested it  and works as expected (i keep saying that it works as expected and then you find that it doesn't work as expected :smile: )

src/lineup/internal/cmds.ts
oltionchampari: Edit: Commenting out this variable produces the above bug. It is used outside the function.
It shouldn't therefore be commented out.

PR #312, opened 2020-02-24
Fixed disappearing of multiple push notifications

### LINKED ISSUE #311: Push notifications don't disappear if there is more than one
* Google Chrome Version 77.0.3865.90

### Steps to reproduce

1. Create more than one push notification (e. g by saving something, create an error, ....) within a short period of time.
 

### Observed behavior
* Push notifications are permanently visible.

### Expected behavior
*  Push notifications disappear after a certain amount of time

Push-Noti1
Push-Noti2

puehringer:
Closes #311.

Previously, the notifications was added to the parent via `afterbegin` flag, such that new notifications where always added before older notifications. The problem is that the notification is then fetched via `parent.lastElementChild!;`, such that always the oldest notification was returned. 

Fetching the child via `parent.firstElementChild!` solves this issue.

### DIFF
--- a/src/notifications.ts
+++ b/src/notifications.ts
@@ -21,7 +21,7 @@ export function pushNotification(level: 'success' | 'info' | 'warning' | 'danger
   <button type="button" class="close" data-dismiss="alert" aria-label="Close"><span aria-hidden="true">&times;</span></button>
   ${msg}</div>`);
 
-  const alert = parent.lastElementChild!;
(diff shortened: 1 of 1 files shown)


lehnerchristian:
I think we should change how the `alert` element is queried from the DOM, so that always the `firstElementChild` is queried. then we always have the latest notification on top

puehringer:
@lehnerchristian Adjusted it to always fetch the first child instead.

PR #314, opened 2020-02-28
Update dependencies

### LINKED ISSUE #313: Update dependencies
See https://github.com/datavisyn/visyn2020/issues/3

We should update all dependencies to the latest accessible versions, except for:
* all webpack dependencies
* all test dependencies
* d3 and bootstrap

It might be necessary to update jQuery and React in a separate step as well.

dvvanessastoiber:
Closes #313 

### Summary
* switch branches in _package.json_
* remove `mkdirp` as it would only be needed for recursion
* update `fork-ts-checker-webpack-plugin` since the formerly used version required Typescript 2 as peerDependency and therefore caused errors
* update `jquery` and `select2`


thinkh:
I've successfully tested Ordino with all branches switched to `update_dependencies_38`. I did not get compile or runtime errors while testing.

Issue #336, opened 2020-03-27
The ViewWrapper removes the HTML if the instance is destroyed and the visible state is kept


lehnerchristian:
## ViewWrapper deletes its own node from the DOM

The ViewWrapper class removes its own HTML node a detail view is mounted in when the instance is destroyed: https://github.com/datavisyn/tdp_core/blob/develop/src/base/ViewWrapper.ts#L209

The problem is that the `node` is only created when the `ViewWrapper` is initialized (see https://github.com/datavisyn/tdp_core/blob/develop/src/base/ViewWrapper.ts#L64), which is done differently in many apps, but generally speaking the ViewWrappers are instantiated when the detail view chooser is built

## Visible state is kept when destroying the instance

A second related problem is that if the view is destroyed the ViewWrappers `visible` state is not set to false, which leads to the bug that if the ViewWrapper instance is destroyed and the main ranking view in an application is clicked that suddenly a detail view is instantiated without choosing a view from the detail view chooser. in the following example `matches && this.visible` is true although visible should be false when the instance was destroyed:
see https://github.com/datavisyn/tdp_core/blob/develop/src/base/ViewWrapper.ts#L298-L300

lehnerchristian:
we agreed that the node should be removed when the `destroy` method is invoked, because that is what is expected. so closing this ticket

PR #337, opened 2020-03-27
Avoid deleting the HTML node when destroying ViewWrapper instance and set visible state correctly

lehnerchristian:
## ViewWrapper deletes its own node from the DOM
for a description, please see the linked ticket

please test this behavior in your applications if it has any side-effects.

but as far as I can see it shouldn't. the current behavior of our detail views is that the ViewWrapper instances are created once and then the detail views are only shown and hidden via a CSS class.

but `node` of those instances should not be deleted, because the `ViewWrapper` class is only instantiated when the detail view chooser is built.

as far as I can see the `destroy` method of the `ViewWrapper` class was never called in the projects I have set up locally

### Edit:

another option would be to insert the HTML node again into the DOM, because we still have the reference to it in `this.node`

## Visible state is kept when destroying the instance

by setting the `visible` attribute of the ViewWrapper instance back to false we avoid recrating a view when we click on the centra LineUp table. see https://github.com/datavisyn/tdp_core/blob/develop/src/base/ViewWrapper.ts#L298-L300

### DIFF
--- a/src/base/ViewWrapper.ts
+++ b/src/base/ViewWrapper.ts
@@ -120,27 +120,20 @@ export default class ViewWrapper extends EventHandler implements IViewProvider {
   }
 
   set visible(visible: boolean) {
-    (async () => { // wrap function body into an immediately invoked function expression which is async to be awaitable
-      const selection = this.inputSelections.get(DEFAULT_SELECTION_NAME);
+    const selection = this.inputSelections.get(DEFAULT_SELECTION_NAME);
 
-      let created = false;
+    if (visible) {
+      this.node.classList.remove('hidden');
+    } else {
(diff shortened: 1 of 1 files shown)


lehnerchristian:
I've close #336 , because we agreed on destroying the ViewWrapper should of course actually delete the HTML node.

# New Description:

The actual problem was that when the `visible` setter is called on a `ViewWrapper` instance the update method is invoked.

The update method checked if the instance is actually available and ` && this.node.getBoundingClientRect().width > 0` which not true in this case.

the size of the node depends on the `active` CSS class of the HTML element the detail views are in, which is set from the outside and currently app specific as it looks to me.

however if this check is present my React view does not render correctly as it is not updated anymore

Issue #331, opened 2020-03-25
Check compatibility of LineUp 4 string filter object with provenance graph


thinkh:
With PR https://github.com/lineupjs/lineupjs/pull/274 the return value of a string filter got formalized as [`IStringFilter`](https://github.com/lineupjs/lineupjs/blob/ca66d3e32684b83c2e7b4e495e88d8d3f70fc97e/src/model/StringColumn.ts#L41-L44). It is now possible to store if the user filtered the string column for missing values. The filter event that tdp_core listens will now return an `IStringFilter`.

We must check the following points:

- [ ] Is the data structure compatible with our tdp_core code (see  #301, #306)
- [ ] Is it possible to open old sessions (created with LineUp v3) with LineUp 4

thinkh:
The filter event of LineUp v3 and v4 before PR https://github.com/lineupjs/lineupjs/pull/274 returns a string or a regexp (e.g., `AB` or `/AB/gm`).

LineUp v4 with PR https://github.com/lineupjs/lineupjs/pull/274 returns the `IStringFilter` object (e.g., `{filter: "AB", filterMissing: false}`, `{filter: /AB/gm, filterMissing: false}`, or `{filter: null, filterMissing: true}`). Hence, we must handle these cases and remain backwards compatibility.

thinkh:
@oltionchampari Please continue my implementation in branch https://github.com/datavisyn/tdp_core/tree/thinkh/331_string-filter-compatibility

- [ ] Extend the tests for `restoreRegExp` in cmds.test.ts so that they support all old AND the new filter interface
- [ ] Refactor the `restoreRegExp` implementation
- [ ] Apply the `restoreRegExp` only to filters in the `setColumnImpl` function

PR #345, opened 2020-03-31
Make LineUp v4 RegExp filter serialization backwards compatible

oltionchampari:
Close datavisyn/tdp_core#331

## Summary 

* Extended tests for `serializeLineUpFilter()` and `restoreRegExp()` functions.
* Refactored the functions to be backwards compatible.

### DIFF
--- a/src/lineup/internal/cmds.ts
+++ b/src/lineup/internal/cmds.ts
@@ -7,7 +7,7 @@ import {IObjectRef, action, meta, cat, op, ProvenanceGraph, ICmdResult} from 'ph
 import {NumberColumn, LocalDataProvider, StackColumn, ScriptColumn, OrdinalColumn, CompositeColumn, Ranking, ISortCriteria, Column, isMapAbleColumn, mappingFunctions} from 'lineupjs';
 import {resolveImmediately} from 'phovea_core/src';
 import i18n from 'phovea_core/src/i18n';
-import {isSerializedFilter, restoreLineUpFilter, serializeLineUpFilter, ILineUpStringFilter} from './cmds/filter';
+import {isSerializedFilter, restoreLineUpFilter, serializeLineUpFilter, isLineUpStringFilter} from './cmds/filter';
 
 
 // used for function calls in the context of tracking or untracking actions in the provenance graph in order to get a consistent defintion of the used strings
@@ -357,7 +357,7 @@ function recordPropertyChange(source: Column | Ranking, provider: LocalDataProvi
     }
 
     if (property === LineUpTrackAndUntrackActions.filter) {
-      newValue = isSerializedFilter(newValue) ? serializeLineUpFilter(newValue as ILineUpStringFilter) : newValue; // serialize possible RegExp object to be properly stored as provenance graph
+      newValue = isLineUpStringFilter(newValue) ? serializeLineUpFilter(newValue) : newValue; // serialize possible RegExp object to be properly stored as provenance graph
     }
 
     if (source instanceof Column) {
--- a/src/lineup/internal/cmds.ts
+++ b/src/lineup/internal/cmds.ts
@@ -462,17 +462,26 @@ export function serializeLineUpFilter(filter: ILineUpStringFilter): ISerializabl
  *
  * @internal
  * @param filter Filter as string or plain object matching the IRegExpFilter
- * @returns {string | RegExp| null} Returns the input string or the restored RegExp object
+ * @returns Returns the input string or the restored RegExp object
  */
-export function restoreRegExp(filter: string | IRegExpFilter): string | RegExp {
-  if (filter === null || !(<IRegExpFilter>filter).isRegExp) {
-    return <string | null>filter;
-  }
+export function restoreRegExp(filter: ILineUpStringFilterValue | IRegExpFilter | ISerializableLineUpFilter, filterMissing = false): ILineUpStringFilter {
 
-  const serializedRegexParser = /^\/(.+)\/(\w+)?$/; // from https://gist.github.com/tenbits/ec7f0155b57b2d61a6cc90ef3d5f8b49
-  const matches = serializedRegexParser.exec((<IRegExpFilter>filter).value);
-  const [_full, regexString, regexFlags] = matches;
-  return new RegExp(regexString, regexFlags);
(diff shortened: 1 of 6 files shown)


tests/lineup/internal/cmds.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  it('filter as null', () => {
+    const lineUpFilter = {filter: null, filterMissing: false};
+    expect(serializeLineUpFilter(lineUpFilter)).toMatchObject({filter: {value: null, isRegExp: false}, filterMissing: false});
+  });
+
+  it('filter as null and use regular expression true', () => {
+    const lineUpFilter = {filter: /null/, filterMissing: false};
+    expect(serializeLineUpFilter(lineUpFilter)).toMatchObject({filter: {value: /null/, isRegExp: true}, filterMissing: false});
```
oltionchampari: What should happen in this case ?
serialize

src/lineup/internal/cmds.ts
```diff
@@ -378,49 +389,98 @@ interface IRegExpFilter {
...
  */
-function restoreRegExp(filter: string | IRegExpFilter): string | RegExp {
-  if (filter === null || !(<IRegExpFilter>filter).isRegExp) {
-    return <string | null>filter;
+export function restoreRegExp(filter: ILineUpStringFilterValue | IRegExpFilter | ISerializableLineUpFilter, filterMissing = false): ILineUpStringFilter {
+  const isIRegExpFilter = ((filter: IRegExpFilter | ISerializableLineUpFilter): filter is IRegExpFilter => filter.hasOwnProperty('isRegExp'));
+  const isISerializableLineUpFilter = (filter: IRegExpFilter | ISerializableLineUpFilter): filter is ISerializableLineUpFilter => filter.hasOwnProperty('filterMissing');
+  if (filter === null || typeof filter === 'string' || Array.isArray(filter)) {
+    return {filter: <string | null>filter, filterMissing};
+  } else if (isIRegExpFilter(filter)) {
```
thinkh: Just a minor code style. I would add a new line before each else block

```suggestion

  } else if (isIRegExpFilter(filter)) {
```

tests/lineup/internal/cmds.test.ts
oltionchampari: After a second inspection i found out that the value  `/null/` is not possible since the change regex filter checkbox reacts only if a value is present.

I added another test for the case when you click the filter missing values and filter is empty.

filter

src/lineup/internal/cmds/filter.ts
```diff
@@ -0,0 +1,126 @@
...
+
+
+/**
+ * This type guard checks if a given parameter has the `filter` property.
+ * Necessary since number columns filter has properties `min`, `max` and no filter property,
+ * @interal
+ *
+ * @param filter Any value that could be a filter
+ */
+export function isSerializedFilter(filter: any): ISerializableLineUpFilter {
```
thinkh, line 62:
```suggestion
export function isSerializedFilter(filter: any): filter is ISerializableLineUpFilter {
```

I would expect that this is a type guard function. And the `return` statement indicates a boolean return value. Please revise.

src/lineup/internal/cmds/filter.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+type ILineUpStringFilterValue = string[] | string | null;
+
+
+/**
+ * This type guard checks if a given parameter has the `filter` property.
+ * Necessary since number columns filter has properties `min`, `max` and no filter property,
+ * @interal
+ *
+ * @param filter Any value that could be a filter
```
thinkh, line 60: `@returns` documentation is missing.

tests/lineup/internal/cmds.test.ts
thinkh: Ok, good. Then we do not need to unit test this.

src/lineup/internal/cmds/filter.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    return restoreLineUpFilter(filter.value, filterMissing);
+
+  } else if (isISerializableLineUpFilter(filter)) {
+    return restoreLineUpFilter(filter.filter, filter.filterMissing);
+
+  }
+
+  throw new Error('Unknown LineUp filter format. Unable to restore the given filter.');
```
thinkh, line 125: I added this error message to prevent cases where the return value of this function is `void`. After playing around with this implementation I got an error with the regexp string:

ordino-lineup-unknown-filter-format

This is the error in the console:

```
filter.ts:125 Uncaught (in promise) Error: Unknown LineUp filter format. Unable to restore the given filter.
    at restoreLineUpFilter (filter.ts:125)
    at restoreLineUpFilter (filter.ts:121)
    at ActionNode.<anonymous> (cmds.ts:227)
    at Generator.next (<anonymous>)
    at fulfilled (tslib.es6.js:71)
```

@oltionchampari Can you check why calling it recursively (starting from line 121) and not returning `void` (without the error) and now we run into the error? In guess is that we still have an uncovered case where the three typeguards above are not working correctly.

My expectation would be that we do not get the error message and one of the typeguards catches the recursive call and returns a valid filter object.

oltionchampari: The issue was that when you go back in the provenance graph the `restoreLineUpFilter()`
function gets called with the old value which gets derived from the Ranking.Thus it hasn't been serialized:

 ```typescript
const fromGraphFilter = {filter: /abc/, filterMissing: false};
   restoreLineUpFilter(fromGraphFilter)
```
I added a test to cover this case and returned the filter when it's already serialized.

src/lineup/internal/cmds/filter.ts
oltionchampari: The function checks if the filter is `ISerializableLineUpFilter | ILineUpStringFilter`, so i added a typeguard that includes both.

thinkh: That fix was only part of the solution. The reason why you had to add `ILineUpStringFilter` to everything was that we were using the function `isSerializedFilter` in `recordPropertyChange` which is incorrect! Instead we must check if the provided filter is a valid `ILineUpStringFilter`. I refactored the code and extracted + tested the type guards.

thinkh:
@oltionchampari I refactored the code again and wrote more type guards and tests. For me it works, but can you please check and test the solution in the browser again. If you do not find any errors or flaws we are ready to merge.

oltionchampari:
It works as expected for me too, i see no problems with it. The code is much better organized now.

Issue #346, opened 2020-03-31
LineUp v4: Ignore live preview events in provenance graph tracking


thinkh:
* Release number or git hash: develop
* Web browser version and OS: independent
* Environment (local or deployed): both

### Steps to reproduce

1. Start an application with the tdp_core#lineupjs4 branch
1. Open _Session_ side panel (provenance graph)
2. Open some LineUp dialogs (e.g., filter, group by, ...) and test the live preview
1. Cancel the dialog

### Observed behavior

![ordino-lineup-live-preview](https://user-images.githubusercontent.com/5851088/78047449-a3d2ea80-7378-11ea-86ef-0c486ff4010f.gif)

All events from the live preview are tracked in the provenance graph.

### Expected behavior

All live preview events should be filtered out. At the end only _confirmed_ changes should be added to the provenance graph.

thinkh:
### Background information

The dialog has been reworked and most of them now provide a live preview. The change was introduced with LineUp PR https://github.com/lineupjs/lineupjs/pull/265. It also includes two new dialog events and listeners, which are fired from the LineUp instance ([initialized in `ALineUp`](https://github.com/lineupjs/lineupjs/blob/6fa424c1782b8d669615a54360c4da5ab82bc575/src/ui/ALineUp.ts#L78-L79)):

```ts
  on(type: typeof ALineUp.EVENT_DIALOG_OPENED, listener: typeof dialogOpened | null): this;
  on(type: typeof ALineUp.EVENT_DIALOG_CLOSED, listener: typeof dialogClosed | null): this;
```

The [`dialogClosed` event listener](https://github.com/lineupjs/lineupjs/blob/6fa424c1782b8d669615a54360c4da5ab82bc575/src/ui/ALineUp.ts#L40) provides an `action` parameter which indicates if the dialog was canceled or confirmed by the user:

```ts
export declare function dialogClosed(dialog: ADialog, action: 'cancel' | 'confirm'): void;
```

A LineUp example can be found in the [dialog demo](https://github.com/lineupjs/lineupjs/blob/6fa424c1782b8d669615a54360c4da5ab82bc575/demo/dialog.html#L53-L59).

### Steps to fix tdp_core

[...]

PR #351, opened 2020-04-03
Ignore live preview events

oltionchampari:
Closes datavisyn/tdp_core#346

## Summary
* To simplify the implementation we decided to integrate  the **lodash** library, instead of writing
our own function to check for deep equality. 
* Refactored the actions that have a **live-preview** to not execute immediately, but instead
to run only when their corresponding dialog has been confirmed.

### Open dialog + SidePanel action
Clicking on an action on the SidePanel when a dialog is open, does not execute the action.

To resolve this issue we extended the blurring effect of the ranking when a dialog is opened to
include the SidePanel.

### Traversing the prov tree node with an open dialog
When clicking on an action in the provenance graph node and a dialog is open in the ranking, the action does not get executed but buffered.

To resolve this issue we close all open dialogs when an action is clicked on, using the LineUp
**DialogManager**.

### Reset all filters

Click on the _reset all filters_ button will push multiple filter actions to the provenance graph, because LineUp is resetting each column individiually and trigger multiple filter events. Hence, going back one state in the provenance graph will only restore a single filter and **not** all filters at once.

ordino-reset-all-filter

A possible solution would be the implementation of compound actions that buffer multiple events and execute them in one go. However, that would require the extension of the provenance tracking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
     "phovea_clue": "github:phovea/phovea_clue#develop",
     "select2": "~4.0.13",
     "select2-bootstrap-theme": "0.1.0-beta.9",
-    "lodash": "^4.17.15"
+    "lodash": "~4.17.15"
   },
   "main": "build/tdp_core.js",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -87,7 +87,8 @@
     "lineupjs": "github:lineupjs/lineupjs#lineup-v4",
     "phovea_clue": "github:phovea/phovea_clue#develop",
     "select2": "~4.0.13",
-    "select2-bootstrap-theme": "0.1.0-beta.9"
+    "select2-bootstrap-theme": "0.1.0-beta.9",
+    "lodash": "^4.17.15"
   },
   "main": "build/tdp_core.js",
   "devDependencies": {
--- a/src/lineup/ARankingView.ts
+++ b/src/lineup/ARankingView.ts
@@ -293,6 +293,11 @@ export abstract class ARankingView extends AView {
       violationChanged: (_: IRule, violation: string) => this.panel.setViolation(violation)
     }));
 
+    // LineUp creates an element with class `lu-backdrop` that fades out all content when a dialog is opened.
+    // Append `lu-backdrop` one level higher so fading effect can be applied also to the sidePanel when a dialog is opened.
+    const luBackdrop = this.node.querySelector('.lu-backdrop');
+    this.node.appendChild(luBackdrop);
+
     this.panel = new LineUpPanelActions(this.provider, this.taggle.ctx, this.options, this.node.ownerDocument);
     this.panel.on(LineUpPanelActions.EVENT_SAVE_NAMED_SET, (_event, order: number[], name: string, description: string, sec: Partial<ISecureItem>) => {
       this.saveNamedSet(order, name, description, sec);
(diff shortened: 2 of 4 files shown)


oltionchampari:
For dialogs that contain multiple actions with live preview, like the visualization dialog,
 i save the last one of each type in a map and when the dialog is confirmed,
 i execute all three adding three entries to the prov graph.

render_type

oltionchampari:
An issue was that when you are buffering the live preview actions and then execute the last one, 
your initial state/value can be the same as the last action value.
Thus pushing an action to the prov graph that doesn't change the state.

To resolve this i save the initial action value and then compare it to the last.
I am using the _lodash isEqual_ as a temporary solution, maybe you have a better suggestion.

thinkh:
> For dialogs that contain multiple actions with live preview, like the visualization dialog, i save the last one of each type in a map and when the dialog is confirmed, i execute all three adding three entries to the prov graph.

That's a good solution for now. At some point, when the provenance graph supports compound actions, we can change it.

> To resolve this i save the initial action value and then compare it to the last. I am using the lodash isEqual as a temporary solution, maybe you have a better suggestion.

I will have a look and think of it once I check the code.

oltionchampari:
> I only tested the functionality and did not have a deeper look into the code yet. In general it works already good, but I the following issues:
> 
> ### Undo restores a wrong state for filter actions
> ordino-lineup-live-preview-last-state
> 
> * This applies to filter dialogs of string, number, and categorical columns
> * Same applies to _sort groups by_ dialog of number columns
> 
> ### No tracking when removing item from hierarchy in side panel
> 1. Categorical column -> More -> _Sort groups by_ **OR** _Group by_
> 2. Ascending
> 3. Confirm -> should be tracked and hierarchy in side panel should be visible
> 4. Remove item from hierarchy in side panel -> action was not tracked
> 
> ordino-lineup-live-preview-last-state
> 
> * This applies to _sort groups by_ and _group by_ dialogs of string, number, and categorical columns
### Update
To resolve the second problem, we decided to fade/blur out the `sidePanel `the same way the lineup columns are when a dialog is opened.

To achieve this result, we append the `lu-backdrop` element one level higher in the node tree.

fade_sidepanel

thinkh:
@oltionchampari Could it be that you forgot to push the change with the change with the `.lu-backdrop`? Because for me the side panel is still not covered. Otherwise I might have made a mistake when resolving the merge conflicts with the last commit. Can you please have a look?

oltionchampari:
> @oltionchampari Could it be that you forgot to push the change with the change with the `.lu-backdrop`? Because for me the side panel is still not covered. Otherwise I might have made a mistake when resolving the merge conflicts with the last commit. Can you please have a look?

I'm sorry, you are right. I didn't push the last commit.

oltionchampari:
### Interactive histogram in side panel
### Reset all filter

The issue was that for the filter action i was always buffering the action. 
Now i execute the action immediately if no dialog has been opened.

### Score column filter
For me this works as expected.

### To discuss
With the current implementation (being dependent on the opening and closing of the dialogs ) 
the playback doesn't work properly when you have a dialog open.

playback

oltionchampari:
So i tested it one more time solution for the **provenance graph + open dialog** bug and it works as expected. All open dialogs close and the buffered actions get reset/emptied.

thinkh:
Thanks for the change and testing. I also tested it again and fixed another edge case, when dialogs are open and a new provenance action is by chaning a parameter from the parameter form.

ordino-close-dialog-on-execute-action


In my opinion we are good to go now. We can open new issues for edge cases that might appear.

Issue #330, opened 2020-03-24
Add Column Dialog in LineUp shows overlapping entries


puehringer:
* Release number or git hash: v9
* Web browser version and OS: 
* Environment (local or deployed): 

### Steps to reproduce

1. Insert many columns into a ranking
2. Use the Sidepanel in 'top' mode (the issue might appear in normal mode too!)

### Observed behavior
The entries are overlapping due to the `max-height` restriction in https://github.com/datavisyn/tdp_core/blob/master/src/styles/_view_lineup.scss#L366. One (untested) solution might be to remove this rule altogether as the actual dialog is properly overflowing.

![image](https://user-images.githubusercontent.com/51900829/77442618-d2861980-6dea-11ea-80f5-e5a65b71e7a7.png)

### Expected behavior
* No overlapping entries with only the scrollbar of the full dialog visible (no subscrolling).

thinkh:
Fixed with the last commit https://github.com/datavisyn/tdp_core/commit/b5fc87f8ef3f3f139811c1b2c76f50efc306cfe5.

![grafik](https://user-images.githubusercontent.com/5851088/77538096-729d7a80-6e9f-11ea-8276-25c71d821ff8.png)

PR #355, opened 2020-04-21
Fix overlapping entries in add column dialog

puehringer:
Fixes datavisyn/tdp_core#330. 
It is already fixed in the Lineup#4 branch (https://github.com/datavisyn/tdp_core/commit/b5fc87f8ef3f3f139811c1b2c76f50efc306cfe5), but we need the change in current develop as well.


thinkh:
Ok, sounds good. 👍

Issue #370, opened 2020-06-03
`AEmbeddedRanking.rebuild()` should return a Promise


dg-datavisyn:
`AEmbeddedRanking.rebuild` should return a Promise, so that the calling Applications knows, when the rebuild is finished. 

https://github.com/datavisyn/tdp_core/blob/1b21fc94f8d1d6ba716ca6b6529cef9dc49b9a82/src/base/AEmbeddedRanking.ts#L166-L170

puehringer:
There is a hook available in the `ARankingView` https://github.com/datavisyn/tdp_core/blob/master/src/lineup/ARankingView.ts#L618-L620 which allows exactly that. The only problem is that when using an `AEmbeddedRanking`, the actual `ARankingView` is nested and accessible via `getInstance()`, not via inheritance, causing the protected function to be inaccessible and not overridable. Therefore, one would have to add the same hook to `AEmbeddedRanking` (and override it in your own ranking) and add the propagation function to the nested `EmbeddedRankingView`. 
```
    class EmbeddedRankingView extends ARankingView implements IEmbeddedRanking {
      ...
      protected builtLineUp(lineup: LocalDataProvider) {
        that.builtLineUp(lineup);
      }
      ...
    }
```
Honestly, I find that very confusing and would prefer a way of unifying the `A[...]Ranking[View].ts`, such that these features can be used universally without propagation/duplication.

thinkh:
Yes, in general I agree with you. @anita-steiner, @dg-datavisyn and me discussed this fact briefly last week. We should refactor and combine the different rankings views and provide one class that can be configured. But since this is a bigger change and requires more time, we should probably aim for the quick fix for now to continue with the other projects. Maybe we can discuss this together with @dg-datavisyn and also the code owner @lehnerchristian in a call.

puehringer:
I just found another issue with this hook approach: the `AEmbeddedRanking.build()` is only executed when using `data+desc+scores` or `data+desc` as `rebuild` parameters. It does not trigger the `builtLineUp` hook when only reloading the scores and/or the data.
Therefore, the solution above can not be used universally.

puehringer:
@dg-datavisyn Does this [PR](https://github.com/datavisyn/tdp_core/pull/597) solve your issues?

dg-datavisyn:
Yes, looks good. Does is also apply to the AEmbeddedRanking?

PR #372, opened 2020-06-08
Added builtLineUp hook propagation in AEmbeddedRanking

puehringer:
_Draft because I don't think we should implement all features like this... see #370 for details_

**Usage:**
After calling `this.rebuild('...')` in your ranking, the overwritten `builtLineUp(lineup)` will be triggered after everything is done.


thinkh:
We close this PR becaues of https://github.com/datavisyn/tdp_core/issues/370#issuecomment-641088494. A better solution is required.

Issue #378, opened 2020-06-17
Persist expand/collapse state for aggregated groups in the provenance graph


thinkh:
* Release number or git hash: develop
* Web browser version and OS: Chrome 83
* Environment (local or deployed): both

### Steps to reproduce

1. Open https://ordino-daily.caleydoapp.org/
1. Open list of genes or cell lines
1. Group by some categorical column
1. Collapse one or multiple groups
1. Reload the page

### Observed behavior

![tdp_core-persist-group-collaps-expand](https://user-images.githubusercontent.com/5851088/84871651-194d7d80-b081-11ea-868c-f46dfd3975f6.gif)

No information is persisted in the provenance graph about the collapse/expand of aggregated groups.

### Expected behavior

Once the collapse/expand information is persisted the ranking should look the same after reloading the page.

thinkh:
Implementation pointers:

* Listen to the `ADataProvider.EVENT_GROUP_AGGREGATION_CHANGED` event from LineUp
* Restore an aggregation with [`ADataProvider.aggregate()`](https://github.com/lineupjs/lineupjs/blob/1f56c5957b50a376c4812a1f7d6de549ff1ced64/src/provider/ADataProvider.ts#L127-L131):
   ```ts
    /** 
	 * @asMemberOf ADataProvider 
	 * @event 
	 */ 
	export declare function aggregate(ranking: Ranking, group: IGroup|IGroup[], value: boolean): void; 
   ```
* Add both listener and aggregate restore to the [cmds.ts](https://github.com/datavisyn/tdp_core/blob/master/src/lineup/internal/cmds.ts)

PR #380, opened 2020-06-22
Persist expand/collapse state for aggregated groups in the provenance graph

oltionchampari:
Depends on lineupjs/lineupjs#353
Closes datavisyn/tdp_core#378

### Summary
* Added clue commands to track the aggregation state of the groups in the provenance graph

### Screenshot
aggregate_column

@thinkh When you have time, to complete this PR, we need to discuss how to handle 2+ levels nested groups. The groups object contains circular references which causes `JSON.stringify()` to fail in phovea_clue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "@types/select2": "4.0.44",
     "d3": "~3.5.17",
     "jquery": "~3.4.1",
-    "lineupjs": "github:lineupjs/lineupjs#sgratzl/aggregationrestore",
+    "lineupjs": "github:lineupjs/lineupjs#develop",
     "phovea_clue": "github:phovea/phovea_clue#develop",
     "select2": "~4.0.13",
     "select2-bootstrap-theme": "0.1.0-beta.9",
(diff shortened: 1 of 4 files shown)


thinkh:
@oltionchampari I had a look into the circular reference problem, but could not come up with a propper solution either. I filed a question issue https://github.com/lineupjs/lineupjs/issues/357 for Sam.

thinkh:
The corresponding LineUp PR https://github.com/lineupjs/lineupjs/pull/364 has been merged and it should work now as expected. 

@oltionchampari Please test the functionality in multiple settings and groupings again. If it works for you as well, please open the PR for review. Thanks.

thinkh:
I still noticed a strange behavior with two aggregation levels. I filed issue https://github.com/lineupjs/lineupjs/issues/365 for that.

oltionchampari:
* One issue i found which is not related to this ticket is explained here https://github.com/lineupjs/lineupjs/issues/373
* The other is that changing the value in the `showTopN` dialog of the aggregation column does not have any effect in the the groups' state (simply does not do anything). It could also be that i'm not using it correctly.
* Other than those two everything works as expected.

Issue #379, opened 2020-06-17
Change of Group By settings for string column is not persisted in the provenance graph


thinkh:
* Release number or git hash: https://github.com/datavisyn/tdp_core/commit/388c3b59fec1d2dc888ce16b70fb2299ae03a3d7
* Web browser version and OS: Chrome 83
* Environment (local or deployed): both

### Steps to reproduce

1. Open https://ordino-daily.caleydoapp.org/
2. Open list of all cell lines
3. Name -> Group By
   * Text starts with …: `A`
4. Confirm dialog
5. Open Group By dialog again
   * Switch to regular expression: `^A`
6. Confirm dialog
5. Open Group By dialog again
   * Switch to by name
6. Confirm dialog

### Observed behavior

![ordino-group-by-session](https://user-images.githubusercontent.com/5851088/84902015-89252d80-b0ac-11ea-8c5e-5b2bbebf75ba.gif)

All the changes are applied correctly to ranking, but they are not stored as new state in the provenance graph. Hence, when replaying it we will not get the correct state that the user has seen before.

### Expected behavior

Switch the group by from starts with to regular expression should be stored in the provenance graph as new state.

### Implementation pointers

Similar to the filters (see #345) there must be some bug in the LineUp cmd.ts.

thinkh:
Please note that the Group By dialog might change slightly after PR ttps://github.com/lineupjs/lineupjs/pull/352. I tested it with the current LineUp v4.0.0 version and after this PR and in both cases we are missing new provenance states.

oltionchampari:
The problem here is that **LineUp** itself does not fire the event `Ranking.EVENT_GROUP_CRITERIA_CHANGED` when the `Group By` dialog's parameters change.

thinkh:
According to Sam we are listening to the wrong criteria for this specific case (see https://github.com/lineupjs/lineupjs/issues/354#issuecomment-647964841):

> why should this event be fired, the group criteria order hasn't changed? it is still a StringColumn. However, the grouping itself changes, and it will fire a corresponding `dirtyOrder` event with the reason `group_dirty` and the `StringColumn` fires `groupingChanged`.

[...]

PR #385, opened 2020-06-26
Track in provenance graph the`groupingChanged` LineUp event

oltionchampari:
Closes datavisyn/tdp_core#379

### Summary
* Added a new cmd `Set Property grouping`.
* The StringColumn's groupBy value can be a RegExp so it was necessary to serialize and deserialize the value similar to the RegExp filter value.
* Tested the StringColumn and NumberColumn implementation. Unfortunately i could not find a DateColumn to test the functionality. @thinkh Could you maybe test this somehow?

The events `StringColumn.EVENT_GROUPING_CHANGED` and `Ranking.EVENT_GROUP_CRITERIA_CHANGED`, 
never intersect apart from when the groupBy dialog is set for the first time.

### Screenshot

grafik

### DIFF
--- a/src/lineup/internal/cmds/filter.ts
+++ b/src/lineup/internal/cmds/filter.ts
@@ -1,3 +1,5 @@
+import {IStringGroupCriteria} from 'lineupjs';
+
 /**
  * Basic LineUp string filter values
  */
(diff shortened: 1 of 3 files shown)


src/lineup/internal/cmds/filter.ts
```diff
@@ -173,3 +175,41 @@ export function restoreLineUpFilter(filter: LineUpStringFilterValue | IRegExpFil
...
+interface ISerializedStringGroupCriteria extends IStringGroupCriteria {
+  values: string[];
+}
+
+/**
+ * Serializes LineUp StringColumn's `Group By` dialog's values, which can contain a RegExp objects to a string.
+ * The return value of this function can be passed to `JSON.stringify()` and stored in the provenance graph.
+ * @param groupBy Value returned from the `Group By` dialog
+ */
+export function serializeGroupByValue(groupBy: IStringGroupCriteria) {
```
thinkh, line 191: Could you please add some unit tests for this function?

src/lineup/internal/cmds/filter.ts
```diff
@@ -173,3 +175,41 @@ export function restoreLineUpFilter(filter: LineUpStringFilterValue | IRegExpFil
...
+    };
+  }
+  return groupBy;
+}
+
+/**
+ * Restores LineUp StringColumn's Group By` dialog's values from the provenance graph and returns an `IStringGroupCriteria`.
+ * @param groupBy Value as saved in the provenance graph.
+ */
+export function restoreGroupByValue(groupBy: ISerializedStringGroupCriteria) {
```
thinkh, line 206: Could you please add some unit tests for this function?

oltionchampari: @thinkh I added test for the serialize and restore functions. You can take a look again when you have time.

PR #396, opened 2020-07-22
Ranking view: make hirarchy of entries

dvdanielamoitzi:
**Summary**
This PR adds the possibility to show the database columns under a specific group:
Screenshot from 2020-07-23 12-14-38

**How to use**
Every `IAdditionalColumnDesc` now accepts a `group?: {parent: string, order?: number};` parameter, similar to the one used in detail views. The parent is simply the name of the group (in this example `Who`, `When`, ...) and the order gives the order of the column within the group (0 comes before 1 ...). 

Also, a ranking can define the order of the actual groups within the dialog (default is alphabetically, with the *unknown* groups still listed as `Database Columns` at first) by passing `databaseColumnGroups: {[key: string]: {order?: number }}` as ranking option. The key is the name (`parent` from above), the value is an options object with an `order` property. 

**To discuss**
Should we indent the groups, such that it is clear that they have the `Database Columns` as parent or should we treat them as they are?


src/lineup/internal/LineUpPanelActions.ts
```diff
@@ -374,6 +394,19 @@ export default class LineUpPanelActions extends EventHandler {
     this.searchBoxProvider.update(items);
   }
 
+  private groupColumnDescs(columnDesc: {text: string, id: string, action: () => void, group: {parent: string, order?: number}}[]): [Map<string, any[]>, ISearchOption[]] {
```
puehringer, line 397: Could we extract the interface for `{text: string, id: string, action: () => void, group: {parent: string, order?: number}}` as it is used above too? I suppose we could also use something like `{text: string} & Pick<IAdditionalColumnDesc, 'group'>` to avoid having to redeclare the group option. 

Also, the return value returns `any`, can we return a valid interface here?

puehringer: @dvdanielamoitzi Thanks for the change!

src/lineup/IARankingViewOptions.ts
```diff
@@ -103,6 +103,16 @@ export interface IARankingViewOptions {
...
 
+  /**
+   * additional information regarding the group of a column
+   */
+  databaseColumnGroups: {[key: string]: {
+    /**
+     * the rank of the current group in the column selector
+     */
+    order?: number
+  }};
```
thinkh, line 114:
```suggestion
  databaseColumnGroups: {
    /**
     * ???
     */
    [key: string]: {
      /**
       * the rank of the current group in the column selector
       */
      order?: number
    }
  };
```

@dvdanielamoitzi I changed the formatting slightly. Does the key just have to be unique or is it used as label? Please add a line of documentation. Thanks.

src/base/interfaces.ts
```diff
@@ -11,9 +11,38 @@ import {AppHeader} from 'phovea_ui';
...
   selectedSubtype?: string;
+  /**
+   * part of the initial ranking by default
+   * @default true
+   */
   initialRanking: boolean;
+  /**
+   * adds the item to a parent group
+   */
+  group?: {
```
thinkh, line 32: I think `group` in the context of the ColumnDesc is a bit misleading, as columns can be grouped and there is also a [GroupColumn](https://github.com/lineupjs/lineupjs/blob/master/src/model/GroupColumn.ts) . Can we rename this to something more precise that it is used for the add column dialog drop down?

thinkh: @dvdanielamoitzi @lehnerchristian @puehringer Any ideas?

puehringer: @thinkh @dvdanielamoitzi @lehnerchristian How about `chooserGroup`, as the `Add Column` dialog is internally called column chooser?

puehringer: @thinkh and I agreed that `chooserGroup` is ok!

PR #402, opened 2020-07-28
Changed CSS selector for main LineUp ranking

puehringer:
Previously, the style of the ranking was applied to *all* child divs of the `lineup` class, which interfered with the way the backdrop is treated now (as same level div).


lehnerchristian:
in the project I've tested the backdrop works again :+1:

PR #409, opened 2020-08-06
Added options forwarding to provenance graph storage manager

puehringer:
Waiting for https://github.com/phovea/phovea_core/pull/196


puehringer:
@thinkh CI only fails because https://github.com/phovea/phovea_core/pull/196 is not merged yet, after merging the build should pass.

PR #416, opened 2020-08-19
Removed Exception constraint in DBMigration upgrade except

puehringer:
Alembic raises exceptions which were previously not caught by the `except Exception` clause, causing the server startup to fail if a specific revision could not be identified (DB is on new branch version, code is on older branch) for example. 
Now, everything is caught properly.


lehnerchristian (commented):
should we also log the exception itself? not sure if it provides useful information

puehringer:
@lehnerchristian `_log.exception` logs the stacktrace as well I think. Is the PR fine for you?

lehnerchristian (approved):

thinkh (approved):

PR #423, opened 2020-09-21
Change Persist Session to Save Session & Persist to Save

oltionchampari:
Part of Caleydo/tdp_bi_bioinfodb#1138

### Summary 

* Rename _Persist Session_ to _Save Session_ to be consistent 
#### before
 grafik

#### after

save_session_after

### Open question
* Think about the persist button icon in the temporary session section
* Renaming ideas: saved, stored, remote session any other?


thinkh (approved):
Looks good to me.

> Think about the persist button icon in the temporary session section

I renamed this tooltip and screen reader text also `Save Session` to be consistent with the dialog that opens on click.

PR #426, opened 2020-09-21
Improve locale strings for better UX

oltionchampari:
Part of Caleydo/tdp_bi_bioinfodb#1138

### Summary 
Fixes the below issues

- [x] UX: Wording _Microsoft Excel (xlsx)_ vs. _Save in Microsoft Excel format (xlsx)_; might be confusing since some user have LibreOffice, etc.

- [x] UX: Wording _Combining Columns_ (current) vs. _Combined Columns_


thinkh (approved):
I removed _Save in_, as it was also used in the export format drop down in the customize download dialog. Other than that it looks fine.

PR #428, opened 2020-09-29
Add option in ATDPApplication to hide the provenance menu (aka `Session Analysis Management`)

oltionchampari:
### Summary
* Added in `ATDPApplication` an option `showProvenanceMenu` to hide the provenance menu.


thinkh (approved):
Works for me! 👍

PR #432, opened 2020-10-07
Refactor stylesheets organiation and imports

dvdanielamoitzi:


src/form/elements/FormMap.ts
```diff
@@ -2,7 +2,6 @@
  * Created by Samuel Gratzl on 08.03.2017.
  */
 
-import 'select2';
```
thinkh, line 5: Please check again, if we need to remove this?

src/form/elements/FormSelect2.ts
```diff
@@ -2,7 +2,6 @@
  * Created by Samuel Gratzl on 08.03.2017.
  */
 
-import 'select2';
```
thinkh, line 5: Please check again, if we need to remove this?

src/scss/components/_view_instant.scss
```diff
@@ -1,4 +1,4 @@
-@import '../vars';
+@import '../abstracts/variables';
```
thinkh, line 1: Why do we need to import this here? We also removed the `@import './vars';` from other files.

thinkh (commented):
@dvdanielamoitzi I found some minor issues. Please check them again. Thanks.

thinkh (approved):

PR #433, opened 2020-10-08
Refactor the add grouped column dialog to support multiple columns

oltionchampari:
Coming from Caleydo/tdp_bi_bioinfodb#1143
### Summary
* Added feature to add multiple database/ support/ combined columns instead of just one at a time

#### before
db_col_before
combined_before

#### after

db_col_after
combined_after

* Changed button text

#### before
adc_before

#### after
button_after


thinkh (approved):
Looks good to me! 👍

PR #435, opened 2020-10-09
Adding the column summary to the exported column header

### LINKED ISSUE #434: Export Utils should also export Lineup Summary
CSV and Excel exported columns should also contain the column summary if set.

dg-datavisyn:
The old version ignored the column summary in the export. A lineup table like this with some summaries
image

 was exported to this

image

There was now way to distinguish the EC50 columns from each other. Now it looks like this:

image

I think it's a minor release because it changes the output of the download

Closes #434


anita-steiner (approved):

thinkh (approved):
Looks good! 👍

PR #437, opened 2020-10-15
Wait for correct LineUp event for aggregation provenance action

thinkh:
Previously, the function `dirtyRankingWaiter()` was used to wait until the provenance action was applied within the LineUp ranking. `dirtyRankingWaiter` waits for the event `Ranking.EVENT_DIRTY_ORDER` and `Ranking.EVENT_ORDER_CHANGED`. However, when applying an aggregation via `provider.setTopNAggregated()` these events are never triggered. Hence, we must listen to the `ADataProvider.EVENT_GROUP_AGGREGATION_CHANGED` event which is triggered after the aggregation has been applied.

See also:

https://github.com/lineupjs/lineupjs/blob/4a97202b8990416ad2ea7c1f542365f1aa69a397/src/provider/ADataProvider.ts#L827-L832


src/lineup/internal/cmds.ts
```diff
@@ -269,9 +269,16 @@ export class LineupTrackingManager {
...
-    const waitForAggregated = LineupTrackingManager.getInstance().dirtyRankingWaiter(ranking);
     LineupTrackingManager.getInstance().ignoreNext = LocalDataProvider.EVENT_GROUP_AGGREGATION_CHANGED;
 
+    let resolver: () => void;
+    const waiter = new Promise<void>((resolve) => resolver = resolve);
+
+    p.on(`${LocalDataProvider.EVENT_GROUP_AGGREGATION_CHANGED}.track`, () => {
+      p.on(`${LocalDataProvider.EVENT_GROUP_AGGREGATION_CHANGED}.track`, null); // disable
+      resolver(); // resolve waiter promise
+    });
```
puehringer, line 280: Wouldn't it be possible to avoid the setting of the resolve callback and do everything *ìnline* instead?

```
const waiter = new Promise<void>((resolve) => {
   p.on(`${LocalDataProvider.EVENT_GROUP_AGGREGATION_CHANGED}.track`, () => {
      p.on(`${LocalDataProvider.EVENT_GROUP_AGGREGATION_CHANGED}.track`, null); // disable
      resolve(); // resolve promise
   });
});
```

thinkh: Yes, good suggestion. I was using the structure of the `dirtyRankingWaiter()` function. I will adapt the code.

puehringer (approved):

PR #444, opened 2020-11-18
Hide `.lu-resize-helper` to fix ranking scrolling

### LINKED ISSUE #442: Ranking scrolling issue
### Steps to reproduce

1. open the ranking
2. scroll down
3. create a lineup filter
4. --> everything works fine
5. remove the filter
6. change the width of any column
7. scroll down
8. create a lineup filter
9. --> the ranking does not update the height, a lot of white space is left on the on the bottom
10. change the width of a column
11. --> the height gets updated again

lineup-scrolling-issue

### Observed behavior
* when creating a lineup filter after changing the width of any column in the ranking the height of the ranking is not updated

thinkh:
Fixes #442

### Summary

The following style forced the ranking footer to be visible:

```scss
.lineup-engine {
  // ...

  footer {
    display: block !important;
  }
```

However, there are two footers:

grafik

The LineUp library hides the `.lu-resize-helper` by default, but with the style above we force it show, which changes the behavior to the one described in #442.

This fix hides the `.lu-resize-helper` resulting in the correct behavior:

ordino-filter

### DIFF
--- a/dist/scss/_view_lineup.scss
+++ b/dist/scss/_view_lineup.scss
@@ -72,7 +72,7 @@
     padding-bottom: 1px; // no idea otherwise scrollbar
   }
 
-  footer {
+  footer:not(.lu-resize-helper) {
     display: block !important;
   }
 }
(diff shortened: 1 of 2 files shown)


dg-datavisyn (approved):
Works for Triage :+1:

anita-steiner (approved):

PR #455, opened 2020-12-18
Remove default SVG style

anita-steiner:
Because it doesn't make sense to define such a default styling for svgs, we should remove that


thinkh (approved):
I tested this PR in Ordino. I needed to move the style to tdp_gene to maintain the axis style (see https://github.com/Caleydo/tdp_gene/pull/149). Other than that it works for me.

PR #457, opened 2020-12-21
Introduce client config

anita-steiner:
needs https://github.com/phovea/generator-phovea/pull/449

### Summary

https://docs.google.com/presentation/d/1k36d_CGs5euOdk73AUNC54mWKzztud17ltzkxhM-KU0/edit#


src/ATDPApplication.ts
```diff
@@ -126,6 +137,28 @@ export abstract class ATDPApplication<T> extends ACLUEWrapper {
     });
   }
 
+  /**
+   * Loads the client config from '/clientConfig.json' and parses it.
+   */
+  public static loadClientConfig<T = any>(): Promise<T | null> {
+    return Ajax.getJSON('/clientConfig.json').catch((e) => {
+      // TODO: Do you want to print an error here, or should it fail silently?
+      console.error('Error parsing clientConfig.json', e);
```
thinkh, line 146: I would log an error and add empty clientConfigs.json to every existing apps. New apps should have an empty _clientConfig.json_ by default.

We should probably remove the TODO marker before merging the PR.

src/ATDPApplication.ts
```diff
@@ -89,7 +93,8 @@ export abstract class ATDPApplication<T> extends ACLUEWrapper {
     showOptionsLink: false,
     showReportBugLink: true,
     showProvenanceMenu: true,
-    enableProvenanceUrlTracking: true
+    enableProvenanceUrlTracking: true,
+    clientConfig: null
```
thinkh, line 97:
```suggestion
    clientConfig: {}
```

Why not initializing an empty client config (see also comment below)?

src/ATDPApplication.ts
```diff
@@ -126,6 +137,28 @@ export abstract class ATDPApplication<T> extends ACLUEWrapper {
...
+    });
+  }
+
+  /**
+   * Loads the client config via `loadClientConfig` and automatically merges it into the options.
+   * @param options Options where the client config should be merged into.
+   */
+  public static initializeClientConfig(options: ITDPOptions): Promise<ITDPOptions> {
+    return ATDPApplication.loadClientConfig().then((parsedConfig) => {
+      options.clientConfig = BaseUtils.mixin(options?.clientConfig || {}, parsedConfig || {});
```
thinkh, line 157:
```suggestion
      options.clientConfig = BaseUtils.mixin(options?.clientConfig, parsedConfig || {});
```

Why not initializing the client config already above? Then we could save this error fallback.

thinkh (changes requested):
I made some minor suggestions for improvements. Please add a description how this works and who is overriding whom.

src/ATDPApplication.ts
thinkh: It is still necessary, because the function might be called from outside then the `options.clientConfig` is `null`.

thinkh (approved):
Thanks for your changes. Looks good now!

PR #472, opened 2021-01-19
Removed isEmpty check of client config before loading it

puehringer:
Removed the `isEmpty` check of the client config, as it will always be empty before loading it (except when someone has passed one already). This causes the client config to be never loaded.


anita-steiner:
let's talk about in detail; it has to check; otherwise all apps without a config will get an error

puehringer:
@anita-steiner We check that in the next line already using `options?.clientConfig || {}`. It is true however that in every application it tries to fetch the clientConfig.json and there is no way to turn that off. Maybe we should add a flag that enables/disables that? 
The problem right now is that I have to pass some initial clientConfig which must not be an empty object or null, even though I don't have any initial values.

anita-steiner (approved):

PR #485, opened 2021-02-16
Fixed clearing of ARankingView with existing selection

puehringer:
Previously, clearing the rankings before the selection causes the selection listener to not find the rankings...


dg-datavisyn:
This seems to resolve my problem from https://github.com/datavisyn/tdp_core/pull/487

thinkh (approved):
I couldn't see a difference in my tests, so I think it works. 😃

Issue #484, opened 2021-02-16
Clearing a ARankingView with a selection does not work


puehringer:
Calling `clear` on a ranking view instance with an existing selection does not work because it clears the rankings *before* the selections, such that the selection clearing fails because the ranking does not exist anymore: 
```
LineUpOrderedRowIndicies.ts:74 Uncaught (in promise) TypeError: Cannot read property 'getOrder' of null
    at Object.<anonymous> (LineUpOrderedRowIndicies.ts:74)
    at Dispatch.apply (dispatch.js:61)
    at fireImpl (AEventDispatcher.js:97)
    at LocalDataProvider.push.LUH7.AEventDispatcher.fireImpl (AEventDispatcher.js:103)
    at LocalDataProvider.push.LUH7.AEventDispatcher.fire (AEventDispatcher.js:74)
    at LocalDataProvider.push.Al6Q.ADataProvider.clearSelection (ADataProvider.js:757)
    at ARankingView.ts:567
```

PR #487, opened 2021-02-19
add check for null

dg-datavisyn:
In triage, this unchecked access to `provider.getFirstRanking()` causes the ranking to crash. For some reason, this seems to be null from time to time. I assume a racing condition, since I can't reproduce properly but it happens often. 

I know that this is only a hotfix, but it'd solve to problem in Triage and should not cause any other problems.


puehringer:
@dg-datavisyn Is this PR opened last week solving a similar issue? https://github.com/datavisyn/tdp_core/pull/485 fixing https://github.com/datavisyn/tdp_core/issues/484

dg-datavisyn:
@puehringer This might also be a solution. I'll check if this also solves my error. Thanks

anita-steiner (approved):

PR #499, opened 2021-04-20
Allow access to Form Instance from FormDialog

### LINKED ISSUE #498: FormDialog should allow access to Form instance
* Release number or git hash: 3916867f2d11b84d7a43e463f6c360fab092ef2f
* Web browser version and OS: Chrome 89
* Environment (local or deployed): both

### Steps to reproduce

1. Have a look at the implementation of a Score Dialog
2. Try to access the elements in the underlying form

### Observed behavior
* In order to get the selected values in a SELECT3 field for example, it is necessary to search for them manually in the HTML using selectors and `dialog.builder.form`. This is an unfortunate workaround because the code will break if there are any changes in the HTML.

### Expected behavior
* Add an accessor to the FormDialog class in order to be able to use functions such as `getElementById()`

dvvanessastoiber:
Closes #498


thinkh (approved):
Looks good to me. I've only added a brief documentation and `null` as default value.

src/form/FormDialog.ts
```diff
@@ -14,6 +14,12 @@ import {select} from 'd3';
 export class FormDialog extends Dialog {
   readonly builder: FormBuilder;
 
+  /**
+   * Contains the `IForm` instance from the FormBuilder build process.
+   * The value is set in `showAsPromise()`. Otherwise this property is `null`.
+   */
+  public formInstance: IForm = null;
```
puehringer, line 21: Maybe change the typing to `IForm | null` to make it clear that it is nullable.

puehringer (changes requested):

src/form/FormDialog.ts
dvvanessastoiber: I've updated the type, thank you for your feedback!

thinkh: Me too 😄

puehringer (approved):

PR #502, opened 2021-04-28
Overwrite the default column widths in lineup to match the new bootstrap 4 font size

oltionchampari:
Coming from Caleydo/tdp_bi_bioinfodb#1217

### Summary 

* see https://github.com/lineupjs/lineupjs/issues/230#issuecomment-601243011

* The default column width in LineUp is _100_ and for strings _200._ 
* This PR changes that to _110_ and _200_ respectively

Note: The equivalent to `100px` width (`14 px` font size) with `16 px` font-size is `113px` (value resulted from testing not calculating)

### Comparison

**Before with Bootstrap 3**

grafik

**Now with Bootstrap 4**

grafik


thinkh (approved):
Looks fine to me and works!

@puehringer Please have a look if you are fine with the implementation.

src/lineup/provider/TDPLocalDataProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  protected instantiateColumn(type: IColumnConstructor, id: string, desc: IColumnDesc, typeFactory: ITypeFactory) {
+    // cache the column width because initializing the `type` class mutates the desc object
+    const columnWidth = desc.width;
+
+    // create a column instance needed for the `isSupportType(col)`
+    const col = new type(id, desc, typeFactory);
+
+    // do nothing if column width is already defined or it is a support type column (e.g., rank, aggregation, selection)
+    if (columnWidth >= 0 || isSupportType(col)) {
```
puehringer, line 19: What if the column itself defines a default (as in https://github.com/lineupjs/lineupjs/blob/develop/src/model/SelectionColumn.ts#L73-L75), but is no support type? Then it would override the default value of the column because we are only looking at the incoming desc, not the "outcoming" width value?
I am not sure if that ever occurs in LineUp, but I think we should check all columns first if this combination occurs anywhere before merging this PR. If it appears it will be much more complicated to achieve properly...

thinkh: Thanks for pointing this edge case out! 👍 My last commit should fix it. Please check again.

grafik

puehringer: Looks good to me.

puehringer (approved):

PR #506, opened 2021-05-14
Explicitly hide help link in header with app option `showHelpLink: false`

thinkh:
Currently the help link cannot be hidden, when the tours manager has found some tours.

This change checks if `options.showHelpLink` from the app customization is set to `false`. If so the link will be hidden.

If registered tours are available, the link with the tours modal can be shown either by setting `options.showHelpLink: true` or keep it `undefined`.

For the implementation the `options` from the constructor paramter have been used to maintain the backwards compability and show the tours modal if the `showHelpLink` is `undefined`. Using the default options would always be `false` and hide the link.


puehringer (approved):
I approve of this change as it is backwards compatible, but would propose to separate the help und tours menu items. In my opinion, they should be two separate buttons (with tours getting its own icon), both having the `show...` in the options. 
This would make it much clearer what is happening. 
We can discuss that in the next Core-Meeting.

thinkh:
I totally agree with you to separate the buttons and avoid the hacky reusage when tours are available. Let's discuss in person how we proceed here.

thinkh:
I close this PR in favor of https://github.com/datavisyn/tdp_core/pull/514 and https://github.com/phovea/phovea_ui/pull/153. The rationales behind this decission are documented in https://github.com/datavisyn/visyn2025/issues/90.

PR #514, opened 2021-06-10
Add tour link instead of reusing the help link

thinkh:
https://github.com/datavisyn/visyn2025/issues/90

Related PR https://github.com/phovea/phovea_ui/pull/153

### Summary

* Add separate tour link that opens the modal dialog (instead of reusing the help link)
* The link is added dynamically to the header when tours are available
* The link can be disabled with the option `showTourLink: false` also when tours are available
* Reorganized promises results in constructor

grafik


src/ATDPApplication.ts
```diff
@@ -112,30 +118,28 @@ export abstract class ATDPApplication<T> extends ACLUEWrapper {
...
-        app: () => this.app
-      });
-
-      BaseUtils.mixin(this.options, {
-        showHelpLink: this.tourManager.hasTours() ? '#' : false // use help button for tours
+    Promise.all([configPromise, i18nPromise])
+      .then(() => {
+        this.build(document.body, {replaceBody: false});
+      })
+      .then(() => {
```
puehringer, line 125: Why exactly the `then` chaining?
Wouldn't it be more readable if we convert this whole thing into an inline async function such that we can actually await things?

```
(async () => {
  await Promise.all([configPromise, i18nPromise]);
  
  this.build(document.body, {replaceBody: false});

  this.tourManager = new TourManager({
  ...
})();
```

puehringer (approved):

Issue #515, opened 2021-06-14
Scrollbar in ranking and tours overlay for non-integer device pixel ratio


thinkh:
* Release number or git hash: Bootstrap v4
* Web browser version and OS: Chrome
* Environment (local or deployed): both

### Steps to reproduce

1. Set the screen scale factor or device pixel ratio of the operating system to a floating point value (e.g., 125% or 1.25)
   ![grafik](https://user-images.githubusercontent.com/5851088/121951281-03899b80-cd5b-11eb-8fc6-0c8cdccc5de4.png)
2. Open a ranking view 
3. Start a tour that contains a highlight (= mask cut-out)

### Observed behavior

The ranking view shows scrollbars for each cell:

![grafik](https://user-images.githubusercontent.com/5851088/120497234-85261480-c3be-11eb-8ba9-0d5f556887be.png)

The screen scaling also have an effect on the position of the tour highlights, for example, for the _Ordino Welcome Tour_.

![grafik](https://user-images.githubusercontent.com/5851088/120501132-a6d4cb00-c3c1-11eb-8c0e-a209e306d5bd.png)

Just for completeness, I tried this on Ordino Daily (with Bootstrap 3) and there it seems to work:

![grafik](https://user-images.githubusercontent.com/5851088/120502057-6590eb00-c3c2-11eb-9961-a962e5e4e2ec.png)

### Expected behavior

* No cell scrollbars should be displayed for the ranking view
* The highlight for the tour should be displayed correctly

PR #516, opened 2021-06-14
Fix scrollbars in cells of ranking and tours overlay for non-integer device pixel ratio

thinkh:
Closes https://github.com/datavisyn/tdp_core/issues/515

### Summary

* Ranking: Increase row height to avoid cell scrollbars for floating-point device pixel ratios
* Adjust the width and height of the tour overlay
* Add window resize listener and update tour stepper

### Explanation

Usually the browser rounds the size for floating-point device pixel ratio and return a size + 1px.
In case of an overlay this additional pixel causes the display of scrollbars of the viewport.
This fix truncates the values instead of rounding it and return 1px less, depending on the device pixel ratio.

### Screenshots

**Ranking**

grafik

**Tours overlay**

grafik

**Resize window with tours overlay**

ordino-tour-resize-window

When shrinking the window the overlay size is still calculated based on the viewport scrollbars. IMO this can only be fixed with a loop or `setTimeout` that is checking a while after rendering the page if the overlay fills the viewport and corrects the size. I tested both implementations, but both still had some edge cases and did not work correctly. That's why removed them and decided for the a simple resize.


src/tour/TourManager.ts
```diff
@@ -397,6 +413,9 @@ export class TourManager {
 
 
   private setUp(tour: Tour, context: any = {}) {
+    window.addEventListener('resize', this.resizeListener, {
```
puehringer, line 416: Instead of `window` we could use `this.backdrop.ownerDocument.defaultView`, such that this functionality would also work in external windows for example.

thinkh: Good idea! I changed it.

PR #518, opened 2021-07-16
Add flag to show/hide clue mode buttons in the header

oltionchampari:
Coming from Caleydo/tdp_bi_bioinfodb#1269

### Summary 

Added flag to show/hide (default true) `clue mode buttons` in the header

grafik


thinkh (approved):
Looks good to me and works as expected.

@puehringer Do you agree with this config setting?

puehringer (approved):

PR #530, opened 2021-08-23
Switch side panel actions to Bootstrap 5.1 and make panel classes public

oltionchampari:
Closes datavisyn/reprovisyn#54
Requires https://github.com/phovea/phovea_ui/pull/158
### Summary

* Fixed a bug with the `data violation button `, the icon was not shown property

data-violation

* Changed color of the side panel left border to secondary

* Removed the data-bs-popper: "static" since it caused a scrollbar to appear when clicking the dropdown 

dropdown-scrollbar

* Added sections to the sidepanel header in order to have button groups with rounded corners instead of just buttons separated by margins with sharp corners

* Added a css class to each action in order to override the default button style in other apps 
To make, the add score button to a primary button, for example, you would in your scss file

```
.lu-adder {
   @extend .btn-primary;
}
```
* Added a faIcon option to the `tdpRankingButton` and `panelTab` extension points

* Removed some unused styles

### Screenshots
---

#### EXPANDED

EXPANDED

---
#### COLLAPSED

COLLAPSED

---
### TOP

TOP


thinkh (commented):
We discussed the style in the core meeting and we agreed that we will create a new icon button style which should be applied to these buttons. The button group should then be replaced by the Bootstrap vertical/horizontal stack with some dividers instead of the gap between the different groups. @dvdanielamoitzi will post the icon button style and mockup very soon.

dvdanielamoitzi:
Here's the link to the figma file: https://www.figma.com/file/GBdWeWg49VO1rvVL0d3cjy/Icon-only-buttons?node-id=0%3A1
and a video of the prototype in case you cannot access the figma file: 
lineup-icon-only-buttons-prototype

puehringer:
@thinkh Yes, I think we already talked about making the panel accessible. Make it `public` and everyone is happy 👍 .

thinkh:
I made LineUp side panel classes public by moving the package `src/lineup/internal/panel` to  `src/lineup/panel` with my last commit.

thinkh (changes requested):
Waiting for phovea/phovea_ui#158 to be finished.

thinkh (approved):
@dvdanielamoitzi Thanks for the changes in https://github.com/phovea/phovea_ui/pull/158/files. The buttons look good to me now.

@puehringer Can you please have a look in specific to my last commit that exposes the add column button options as part of the `IARankingViewOptions`. I've added it "as just another option", but I think we should refactor the whole options at one point and maybe introduce a panel or toolbar configuration at one point which gathers and groups all of the different settings that a currently more or less a plain list. But I also think that this refactoring is beyond this PR. Long story short: Are you fine with this PR or do you suggest any changes?

thinkh (changes requested):
I need to withdraw my approval due to a bug in the z-index in the add column dialog. Probably the z-index has changed due to the new nesting inside the `.panel-toolbar`.

**Old behavior**

The toolbar buttons overlay the add column blocker hence, they are hoverable/clickable and close the overlay.

ordino-side-panel-button-z-index



**New behavior (now)**


The toolbar buttons are below the the add column blocker and not hoverable/clickable anymore.


ordino-side-panel-button-z-index-2

@oltionchampari Please take a look at the z-index of the add column blocker and the `.panel-toolbar`. Thanks.

puehringer:
Works great for me, except that the top-mode sets https://github.com/datavisyn/tdp_core/blob/08163b8c033368436ea76f84656765d0274fe467/src/scss/components/_view_lineup.scss#L455 which causes overlap issues:

image

Without the z-index everything works, would it be a good idea to remove it with this PR to test only once?

thinkh:
@dvvanessastoiber Please take over the tasks from @oltionchampari and have a look at the following points in Ordino:

**Prerequisites**
Checkout the following branches
- This branch
- https://github.com/phovea/phovea_ui/pull/158
- https://github.com/datavisyn/tdp_uploaded_data/pull/100
- https://github.com/Caleydo/tourdino/pull/80

**Tasks**

- [ ] Check if buttons from tdp_uploaded_data and tourdino are displayed correctly
- [ ] Ordino rankings (especially in tdp_gene and tdp_publicdb) and add `panelAddColumnBtnOptions` to the following ranking options:
   ```ts
	  {
	    // ... existing ranking options
	    panelAddColumnBtnOptions: {
	      btnColor: 'btn-primary'
	    }
	  }
   ```
- [ ] `z-index` of add column dialog see https://github.com/datavisyn/tdp_core/pull/530#pullrequestreview-753288082

puehringer (approved):
Works for me!

dvvanessastoiber:
@thinkh I have integrated the requested changes. Please test the PR again, taking the following PRs into account:

- https://github.com/Caleydo/tdp_gene/pull/160
- https://github.com/Caleydo/tdp_publicdb/pull/159

dvvanessastoiber:
@thinkh Thank you for fixing the z-index! I have tested it again and it works for me :+1:

thinkh (approved):
Since this PR looks good to everyone now, I'm going to merge it. Thanks to everybody involved for the joint effort! 👍

PR #547, opened 2021-09-06
Fix params forms in ChooserProxyView

thinkh:
### Summary

* Before the parameter form was wrapped because the `col-sm-auto` was missing

**Before**

grafik

**After**

grafik


dvvanessastoiber (approved):
Works in my tests, thank you for the changes :+1:

puehringer (approved):

PR #549, opened 2021-09-14
General Vis Component v1

dvzacharycutler:
General vis components, as demoed and discussed. All new files exist in the generalVis folder in tdp_core, as well as one GeneralVisWrapper file added to the lineup folder. 

Current plotly components include Scatterplots, violin plots, strip plots, bar charts, and PCP. 

To make generic, each plot type inherits from a GeneralPlot interface. Plot classes need to have a function with defines their starting heuristic if necessary, as well as a function that returns the necessary traces for Plotly. 

To make a generic sidebar, there is also a large object with contains a key for each possible form object. The GeneralPlot classes decide which form objects they want displayed by returning a list of keys. 

Quite a few Enums and types for code clarity are in the types/generalTypes file


package.json
```diff
@@ -80,19 +80,25 @@
     "db-migration:list": "npm run db-migration:base list --"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.36",
+    "@fortawesome/free-solid-svg-icons": "^5.15.4",
+    "@fortawesome/react-fontawesome": "^0.1.15",
```
thinkh, line 85: Why do we need these fontawesome dependencies? Fontawesome comes already with phovea_ui and can be used by adding the classes to the element (e.g., `<i class="fas fa-question"></i>`).

src/generalVis/components/GeneralHome.tsx
```diff
@@ -0,0 +1,265 @@
...
+    const updateSelectedCatCols = (s: string[]) => setSelectedCatCols(s);
+    const updateSelectedNumCols = (s: string[]) => setSelectedNumCols(s);
+
+    const updateBarDisplayType = (s: EBarDisplayType) => setBarDisplayType(s);
+    const updateBarGroupType = (s: EBarGroupingType) => setBarGroupType(s);
+    const updateBarDirection = (s: EBarDirection) => setBarDirection(s);
+
+    const shapeScale = useMemo(() => {
+        return shape ?
+            scale.ordinal<string>().domain(d3.set(shape.vals.map((v) => v.val)).values()).range(['circle', 'square', 'triangle-up', 'star'])
```
thinkh, line 54:
```suggestion
            d3.scale.ordinal<string>().domain(new Set(shape.vals.map((v) => v.val)).values()).range(['circle', 'square', 'triangle-up', 'star'])
```

Can we use a regular `Set` here and avoid D3?

src/generalVis/components/GeneralHome.tsx
```diff
@@ -0,0 +1,265 @@
+import d3 from 'd3';
+import {scale} from 'd3';
```
thinkh, line 2:
```suggestion
```

Use `d3.scale` below

src/generalVis/components/GeneralHome.tsx
```diff
@@ -0,0 +1,265 @@
...
+    }, [bubbleSize]);
+
+    const opacityScale = useMemo(() => {
+        return opacity ?
+            scale.linear().domain([0, d3.max(opacity.vals.map((v) => v.val))]).range([0, 1])
+            : null;
+    }, [opacity]);
+
+    const colorScale = useMemo(() => {
+        return scale.ordinal().range(['#337ab7', '#ec6836', '#75c4c2', '#e9d36c', '#24b466', '#e891ae', '#db933c', '#b08aa6', '#8a6044', '#7b7b7b']);
```
thinkh, line 71: @dvdanielamoitzi will add the color scale to phovea_ui. Once available, please repelace this array here.

thinkh (changes requested):
While testing your PR, I saw that you added some fontawesome dependencies. Are they necessary? Please fix at least this before we merge this PR. If we want to keep these dependencies, we should ask @puehringer before adding them.

Reminder to address the following tasks in a next PR:
- [ ] Extract language strings using i18next

thinkh (approved):
We agreed that we merge the changes into the `general-purpose-vis` feature branch and do further refactorings in later PRs.

PR #553, opened 2021-09-20
General Purpose Vis Bar Chart and Various Improvements

dvzacharycutler:
**Features**

Removed Plotly toolbar, added the functionality for switching between brushes as buttons. Also added opacity slider here. Not sure how this should work with non scatterplots, and I just left it for now. Hiding it makes some sense to me, as selection/opacity changes dont make a ton of sense right now in the other plots, but could be useful at some point. 

Screenshot from 2021-09-20 17-31-45

Sidepanel improvements, including 

- Making Numerical and Categorical dropdowns multi selectable without closing
- Always show advanced options
- Resort advanced options, removing those that dont make sense for different visualizations
- Removed the many buttons to switch between vis, made it just a dropdown
- 
Screenshot from 2021-09-20 17-35-05

Label improvements for scatter plots. Similar label issues still need to be handled in other plots in a future ticket. 

Screenshot from 2021-09-20 17-35-56

Scatterplot coloring by numerical attribute. Added both sequential and divergent coloring. This works okay, but the divergent colors need to be changed to have a less white color in the middle. Additionally, usability wise it is not clear how to use this feature, as you first need to select a numerical column to activate the option to switch to divergent/sequential via the sidebar. 

Screenshot from 2021-09-20 17-38-41

Bar Chart is functional, though with some usability problems. It is not clear that the numerical columns do nothing, or that the categorical columns after the first also do nothing. Also need to shorten labels, similar to scatterplot. Grouping/small multiples works. 

Screenshot from 2021-09-20 17-40-48


dg-datavisyn (changes requested):
Already looks great 👍 


Some issues I noticed:

## Scatter Chart

- [x] Scatterplot color by numeric value looks strange. I colored by `pvalue` which is also the y-axis. So I expected a smooth gradient along the y axis
image
- [x] Please add tooltips to the selection type changes (brush, lasso) 
image
- [x] Could we use a bounding box icon for brush selection and a lasso icon for lasso selection?
- [x] Selection is really slow (maybe an extra ticket)
- [x] Opacity change is slow, maybe add a debounce to the slider
- [x] Could we add little color ranges to the `Divergent` and `Sequential` buttons so that is clear to the user what is meant here, like this (but horizontal): 
image
- [ ] Please add a tooltip to the `Divergent` and `Sequential` buttons as well as to the `Filter In | Filter Out | Clear`
- [ ] There is no legend for the color
- [x] :bug: `Filter In | Filter Out | Clear` buttons aren't working (Related question: Does this make sense if the component is used without lineup?)

## Bar Chart

- [x] When I switch to bar chart it takes an integer attribute on the x-axis but it is not listed: 
image
- [ ] The numerical columns are not used in the bar chart
- [x] When I switch from vertical to horizontal the axes labels are not switched: 
image

dvzacharycutler:
Fixed most of the mentioned problems, which were primarily bug fixes. The only new stuff that wasn't a bug fix is:

1. Selection is much faster. Simplified the way I stored selection to move an n^2 selection time to n. Remaining improvements could be made to lineup itself to increase selection speed. 
2. Added color scale beneath the sequential/divergent buttons
Screenshot from 2021-09-22 15-43-06

thinkh:
We agreed that functionality-wise the PR is done and code-wise it will be changed/improved with the refactor in the next PR. Hence, I'll continue and merge this PR.

PR #555, opened 2021-09-22
Selection speed increase by replacing `LineupUtils.array_diff` with lodash `difference`

dvzacharycutler:
Small change to some internal lineup logic to make selecting many points faster. Also a comment change to better explain a function that confused me. Only noticeable when selecting > 10000 or so points, which is way I'm basing it into the generic vis branch.


thinkh (approved):
Thanks for your improvement and well done! 👍 It looks good to me and works as expected.

src/lineup/internal/LineUpSelectionHelper.ts
```diff
@@ -53,8 +53,8 @@ export class LineUpSelectionHelper extends EventHandler {
 
   private onMultiSelectionChanged(indices: number[]) {
     // compute the difference
-    const diffAdded = LineupUtils.array_diff(indices, this.orderedSelectedIndices);
-    const diffRemoved = LineupUtils.array_diff(this.orderedSelectedIndices, indices);
+    const diffAdded = [...LineupUtils.set_diff(new Set(indices), new Set(this.orderedSelectedIndices))];
+    const diffRemoved = [...LineupUtils.set_diff(new Set(this.orderedSelectedIndices), new Set(indices))];
```
thinkh, line 57: After looking at the code again, I have a suggestion. We could make the `array_diff` function "backward-compatible" so that every usage benefits of the performance improvement. All we need to do is to modify the function like this:

```ts
  static array_diff<T>(array1: T[], array2: T[]) {
    return [...LineupUtils.set_diff(new Set(array1), new Set(array2))];
  }
```

Afterwards we can remove the todo from the JSDoc comment and revert the changes here:

```suggestion
    const diffAdded = LineupUtils.array_diff(indices, this.orderedSelectedIndices);
    const diffRemoved = LineupUtils.array_diff(this.orderedSelectedIndices, indices);
```

P.S. I cannot attach the suggestion to the function, because Github allows suggestions only to changed lines.

thinkh (changes requested):

src/lineup/internal/LineUpSelectionHelper.ts
puehringer: We thought about doing something similar to that, the only question that remains is the # of elements where the `Set` approach actually improves performance. Imagine if both arrays are rather small, then the overhead of creating two sets and an `Array.from` will be larger than the gained performance increase from `Set`. 
Generally, it would be a good idea to look at specific performance benchmarks, but that would probably be overkill. 
In short: I think it's best to keep `array_diff` as is, and create a `set_diff(array1: T[] | Set<T>, array2: T[] | Set<T>): Set<T>`, and then create a Set from the parameters if they are arrays. With that, the `new Set(...)` moves inside the function, and it is still clear what the `set_diff` does.

thinkh: Thanks for the heads up! I agree with your suggestions and have one more: Should we rename the functions to `arrayDiff` and `setDiff` (eventhough the last one might be slightly confusing)?

thinkh: @dvzacharycutler Can you remove `array_diff()` and `set_diff()` (= delete the source code) and refactor it to [Lodash `difference()`](https://lodash.com/docs/4.17.15#difference)? Please test the performance of the lodash implementation.

@dvdanielamoitzi After Zach's refactor, please refactor the usage of `array_diff()` in Triage.

thinkh: @dvzacharycutler I did the change. Nothing more to do from your side.

thinkh (approved):
I removed all occurences of `array_diff` and `set_diff` ([as we agreed](https://github.com/datavisyn/tdp_core/pull/555#discussion_r772241697)) and replaced it with lodash `difference`.

@dvdanielamoitzi Please refactor the usage of `array_diff()` in Triage. It's basically a search for `array_diff` and replace with `difference` + updating the corresponding imports.

thinkh:
@puehringer You can merge this PR if you agree with the changes. Thanks.

PR #556, opened 2021-10-04
Refactor general purpose vis

dvzacharycutler:
Full refactor of the general vis component, moving to individual React components for every plot, as well as for every sidebar option. 

As discussed, I put an emphasis on making every component capable of existing as a standalone components, mostly via optional props. 

The state of the general vis can be defined entirely via the top level visConfig state, which is serializeable, to make it easy for future provenance/pre-setting of the vis. 

One problem that definitely still exists is the complexity of the utils files, specifically the complexity of the code to actually create the plotly data objects. vis/bar/utils is a good example of this, as it can turn into a series of if/else depending on config options with large plotly objects being created within making for rather difficult to understand code. 

**Functionality changes**

There aren't too many functionality changes here, mostly quality of life stuff. scatterplots have drop lines on hover, you can move x/y columns in numerical now via drag and drop, numerical color button group now appears below the color select.


src/vis/Vis.tsx
```diff
@@ -0,0 +1,116 @@
...
+    colors?: string[];
+    shapes?: string[];
+    selectionCallback?: (s: number[]) => void;
+    filterCallback?: (s: string) => void;
+}
+
+export function Vis({
+    columns,
+    selected = {},
+    colors = ['#337ab7', '#ec6836', '#75c4c2', '#e9d36c', '#24b466', '#e891ae', '#db933c', '#b08aa6', '#8a6044', '#7b7b7b'],
```
thinkh, line 28: Please retrieve the colors using the `getCssValue()` function introduced with PR https://github.com/phovea/phovea_ui/pull/161.

src/vis/bar/BarVis.tsx
```diff
@@ -0,0 +1,212 @@
...
+        return beautifyLayout(traces, layout);
+    }, [traces, config.groupType]);
+
+    return (
+        <div className="d-flex flex-row w-100 h-100">
+            <div className="position-relative d-flex justify-content-center align-items-center flex-grow-1">
+                {mergedExtensions.prePlot}
+                {traces.plots.length > 0 ?
+                    (<Plot
+                        divId={'plotlyDiv'}
```
thinkh, line 129: If this id is added as `id` attribute to the DOM element then it must be unique, since it is most likely that multiple visualizations used on the same page. Please also change this in the other visualizations.

src/vis/bar/BarVis.tsx
```diff
@@ -0,0 +1,212 @@
...
+                        message={traces.errorMessage} />)
+                }
+                {mergedExtensions.postPlot}
+
+            </div>
+            <div className="position-relative h-100 flex-shrink-1 bg-light">
+                <button className="btn btn-primary-outline" type="button" data-bs-toggle="collapse" data-bs-target="#generalVisBurgerMenu" aria-expanded="true" aria-controls="generalVisBurgerMenu">
+                    <i className="fas fa-bars"/>
+                </button>
+                <div className="collapse show collapse-horizontal" id="generalVisBurgerMenu">
```
thinkh, line 160: This `id` attribute must be unique as well. Please udpate the other occurences accordingly.

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
+import {merge} from 'lodash';
+import {CategoricalColumn, ColumnInfo, EColumnTypes, ESupportedPlotlyVis, IVisConfig, NumericalColumn, Scales} from '../interfaces';
+import {PlotlyInfo, PlotlyData} from '../interfaces';
+import {getCol} from '../sidebar/utils';
+
+export enum EBarDisplayType {
+    DEFAULT = 'Default',
+    NORMALIZED = 'Normalized',
```
thinkh, line 8: Is the value of the enum property used as label for the select? If yes, we would need to extract it as language string and use i18n.

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+                            yaxis: counter === 1 ? 'y' : 'y' + counter,
+                            showlegend: counter === 1 ? true : false,
+                            type: 'bar',
+                            name: uniqueGroup,
+                            marker: {
+                                color: scales.color(uniqueGroup),
+                            }
+                        },
+                        xLabel: vertFlag ? catCurr.info.name : normalizedFlag ? 'Percent of Total' : 'Count',
+                        yLabel: vertFlag ? normalizedFlag ? 'Percent of Total' : 'Count' : catCurr.info.name
```
thinkh, line 132: Here are also some language strings

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    const catCols: CategoricalColumn[] = columns.filter((c) => config.catColumnsSelected.filter((d) => c.info.id === d.id).length > 0 && c.type === EColumnTypes.CATEGORICAL) as CategoricalColumn[];
+    const vertFlag = config.direction === EBarDirection.VERTICAL;
+    const normalizedFlag = config.display === EBarDisplayType.NORMALIZED;
+    const plots: PlotlyData[] = [];
+
+    if(catCols.length > 0) {
+        const catCurr = catCols[0];
+        if(config.group && config.multiples) {
```
thinkh, line 93: Maybe extract the different if-else blocks into separate functions to make the if/else here more readable.

```suggestion
        if(config.group && config.multiples) {
        	plots = lotsForGroupAndMultiple(columns, config);
    	} else if(...) {
```

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    if(!config.catColumnsSelected) {
+        return {
+            plots: [],
+            legendPlots: [],
+            rows: 0,
+            cols: 0,
+            errorMessage: 'To create a Bar Chart, please select at least 1 categorical column.',
+            formList: ['groupBy', 'barMultiplesBy', 'barDirection', 'barGroupType', 'barNormalized']
+        };
```
thinkh, line 83:
```suggestion
		throw new Error('To create a Bar Chart, please select at least 1 categorical column.');
```

In case of error it would make sense to throw errors and handle them in a parent component. It might be also useful to create a custom `VisError` class (somewhere) that can provide custom properties if necessary:

```ts
export class VisError extends Error {

}
```

If you wanna keep the object like you have it: do we still need the `fromList` property?

Please address for the other visualizations accordingly.

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    return merged;
+}
+
+export function createBarTraces(
+    columns: (NumericalColumn | CategoricalColumn)[],
+    config: IBarConfig,
+    scales: Scales,
+): PlotlyInfo {
+    let counter = 1;
```
thinkh, line 73: What does this count? Instances? Plots? Bars? It's unclear without reading the whole code.
Please rename to something more specific.

src/vis/bar/utils.ts
```diff
@@ -0,0 +1,240 @@
...
+    const rows = Math.ceil(Math.sqrt(counter - 1));
+    const cols = Math.ceil((counter - 1) / rows);
+
+    return {
+        plots,
+        legendPlots: [],
+        rows,
+        cols,
+        errorMessage: 'To create a Bar Chart, please select at least 1 categorical column.',
+        formList: ['groupBy', 'barMultiplesBy', 'barDirection', 'barGroupType', 'barNormalized']
```
thinkh, line 238:
```suggestion
```

Can we remove this properties if we throw errors (as suggested above)?

src/vis/interfaces.ts
```diff
@@ -0,0 +1,75 @@
...
+    IN = 'Filter In',
+    OUT = 'Filter Out',
+    CLEAR = 'Clear',
+}
+
+export type IVisConfig = IScatterConfig | IViolinConfig | IBarConfig | IStripConfig | IPCPConfig;
+
+export interface NumericalColumn {
+    info: ColumnInfo;
+    vals: {id: number, val: number}[];
```
thinkh, line 39:
```suggestion
    values: {id: number, val: number}[];
```

src/vis/interfaces.ts
```diff
@@ -0,0 +1,75 @@
...
+    type: EColumnTypes.NUMERICAL;
+    selectedForMultiples: boolean;
+}
+
+export interface CategoricalColumn {
+    info: ColumnInfo;
+    colors: string[];
+    vals: {id: number, val: string}[];
```
thinkh, line 47:
```suggestion
    values: {id: number, val: string}[];
```

src/vis/layoutUtils.ts
```diff
@@ -0,0 +1,126 @@
+import {PlotlyInfo} from './interfaces';
+
+/**
+ * Truncate long texts (e.g., to use as axes title)
+ * @param text Input text to be truncated
+ * @param maxLength Maximum text length (default: 50)
+ */
+function truncateText(text: string, maxLength = 50) {
+    return (text.length > maxLength) ? `${text.substr(0, maxLength)}...` : text;
```
thinkh, line 9:
```suggestion
    return (text.length > maxLength) ? `${text.substr(0, maxLength)}&hellip;` : text;
```

Could we use `hellip;` which renders the typographically correct dots?

src/vis/layoutUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+export function beautifyLayout(traces: PlotlyInfo, layout: any) {
```
thinkh, line 12: Please add a function comment, because it is uclear what beautify layout does or means in this context.

src/vis/scatter/utils.ts
```diff
@@ -0,0 +1,257 @@
...
+        min = d3.min((getCol(columns, config.color) as NumericalColumn).vals.map((v) => +v.val).filter((v) => v !== null)),
+        max = d3.max((getCol(columns, config.color) as NumericalColumn).vals.map((v) => +v.val).filter((v) => v !== null));
+    }
+
+    const numericalColorScale = config.color ?
+                d3.scale.linear<string, number>()
+                    .domain([max,
+                        (max + min) / 2,
+                        min])
+                    .range(config.numColorScaleType === ENumericalColorScaleType.SEQUENTIAL ? ['#002245', '#5c84af', '#cff6ff'] : ['#337ab7','#d3d3d3', '#ec6836'])
```
thinkh, line 101: Please retrieve the colors from PR https://github.com/phovea/phovea_ui/pull/161/.

thinkh (changes requested):
Thanks for your refactoring, it looks already very good! 👍

I had a quick look and saw the following points that should be adressed:

- [ ] Connect this PR to the issue in Zenhub
- [x] When renaming a file/folder please be sure to remove the old file/folder from the `dist` directory. Currently, there is still a `dist/generalVis` folder
- [x] Retrieve colors using the `getCssValue()` function that was introduced with PR https://github.com/phovea/phovea_ui/pull/161/
- [ ] Maybe refactor error handling and throw/catch errors
- [x] Make `id` attributes unique, due to multiple Vis instances on the same page
- [ ] Add support for `data-testid` (ask @dvvanessastoiber for further details)
- [ ] Extract language strings and use i18n (could also be done in a follow-up PR)

See also the comments below.

thinkh:
Open points:

- [ ] Add option flag to enable/disable the general vis in the ARankingView options
- [ ] Add a message that this feature is experimental

src/vis/violin/utils.ts
```diff
@@ -0,0 +1,142 @@
...
+            plots: [],
+            legendPlots: [],
+            rows: 0,
+            cols: 0,
+            errorMessage: 'To create a Violin plot, please select at least 1 numerical column.',
+            formList: ['violinOverlay']
+        };
+    }
+
+    const numCols: NumericalColumn[] = columns.filter((c) => config.numColumnsSelected.filter((d) => c.info.id === d.id).length > 0 && EColumnTypes.NUMERICAL) as NumericalColumn[];
```
puehringer, line 57: I am not sure if I understand the filter condition (also the one below), i.e. what does the `&& EColumnTypes.NUMERICAL` do? Shouldn't there be a `&& c.something === EColumnTypes.NUMERICAL` instead?

puehringer: I just found it: it should probably look like:
```typescript
const numCols: NumericalColumn[] = columns.filter((c) => config.numColumnsSelected.filter((d) => c.info.id === d.id).length > 0 && c.type === EColumnTypes.NUMERICAL) as NumericalColumn[];
```

Also, instead of checking the length if you have any matches, use `.some(` as it returns early (i.e. when it found a single match):

```typescript
const numCols: NumericalColumn[] = columns.filter((c) => config.numColumnsSelected.some((d) => c.info.id === d.id) && c.type === EColumnTypes.NUMERICAL) as NumericalColumn[];
```

puehringer:
Could you add some documentation on how to use this package as a standalone feature?
I.e. add a `README.md` in the `generalVis` package outlining the contents, what can be done in what way, etc. This can be a fairly thorough document, but it makes our lives much easier. I have done something similar for the `UseBootstrap` hooks [here](https://github.com/datavisyn/tdp_core/blob/master/src/hooks/useBootstrap.md).

Also @dvzacharycutler  @thinkh, can we rename the package to `vis`? I don't think that `generalVis` is a nice name, especially since we are working on using named export soon, i.e. by allowing `import {...} from 'tdp_core/generalVis'`. Here, `import {...} from 'tdp_core/vis'` looks much nicer.

src/vis/layoutUtils.ts
dvzacharycutler: done

src/vis/layoutUtils.ts
dvzacharycutler: done

src/vis/scatter/utils.ts
dvzacharycutler: done except for d3d3d3, since that is not one of our colors

src/vis/Vis.tsx
dvzacharycutler: done

src/vis/bar/BarVis.tsx
dvzacharycutler: done

src/vis/bar/BarVis.tsx
dvzacharycutler: done

src/vis/bar/utils.ts
dvzacharycutler: agreed, though i'd love to find some way in the future to reduce the crazy code duplication here, but plotly makes it hard. Separated into different functions for now.

src/vis/bar/utils.ts
dvzacharycutler: changed to plotCounter in all files

src/vis/bar/utils.ts
dvzacharycutler: even without the errors this was not necessary any more, removed it

src/vis/violin/utils.ts
dvzacharycutler: good catch, yes that was a bug. and the some feature is great to know about! Changed a few other places to also use .some

src/vis/interfaces.ts
dvzacharycutler: done

src/vis/interfaces.ts
dvzacharycutler: done

src/vis/bar/utils.ts
dvzacharycutler: I agree with this, though do worry slightly about making parent components deal with the errors. But I think as long as there is a sensible default for what happens that can be overridden via props, this is a great idea. But it will be left to a future PR

src/vis/bar/utils.ts
dvzacharycutler: Since I imagine the way we do this will change in the near future, leaving any implementation of i18n to future PRs.

src/vis/bar/utils.ts
dvzacharycutler: Since I imagine the way we do this will change in the near future, leaving any implementation of i18n to future PRs.

puehringer:
@dvzacharycutler 
* Lineup column names can contain arbitrary HTML: (I cannot include a screenshot for some reason...)
Screenshot 2021-12-10 081552

* The selection can only be cleared when a selection is active. I.e. to reproduce select something in a scatterplot, click filter in, then double click somewhere in the plot to remove the selection, then click clear --> nothing happens. Selecting something and then clicking clear works though. Clear only works when a selection is active.

dvzacharycutler:
@puehringer thanks for the great feedback, some of those bugs I never would have found or noticed without it. Implemented/fixed everything you mentioned except for the react-split. I certainly agree that would be a great addition, but that is what we are doing in Ordino 2.0 already, so maybe better to just wait?

thinkh (approved):
@dvzacharycutler Thanks for this huge refactoring and the latest changes. To me it looks good now. After merging https://github.com/datavisyn/tdp_core/pull/600, we can merge the general vis can be merged into develop. 

@puehringer Please have a quick last look and approve this PR. Please merge this PR afterwards. Thanks.

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -54,11 +55,11 @@ export class GeneralVisWrapper extends EventHandler {
...
-    selectCallback(selected: number[]) {
-        const r = Range.list(selected);
-        const id = IDTypeManager.getInstance().resolveIdType(this.view.itemIDType.id);
-        this.view.selectionHelper.setGeneralVisSelection({idtype: id, range: r});
-    }
+    // selectCallback(selected: number[]) {
+    //     const r = Range.list(selected);
+    //     const id = IDTypeManager.getInstance().resolveIdType(this.view.itemIDType.id);
+    //     this.view.selectionHelper.setGeneralVisSelection({idtype: id, range: r});
+    // }
```
puehringer, line 62: @dvzacharycutler You can probably remove this as well now.

dvzacharycutler:
@puehringer regarding the normalization problems were running into, I believe it is a problem with the lineup columns getValue() function, as the vis component is the only place that actually accesses the values this way, and they are already normalized coming out of lineup. So our options seem to be to either make a change to lineup or go back to getting the score values via the accessor directly.

PR #557, opened 2021-10-06
Added `useAsyncFunction` as parametric `useAsync` version

puehringer:
The existing `useAsync` does only allow async functions without parameters. 
This new hook allows the usage of paramters, with `useAsync` being the special case without any parameters. 

TODO @thinkh: The only non-backwards compatible thing here is the order of the generics, and the second parameter. Maybe we refactor all usages (should not be too many) to the new version?


puehringer:
@thinkh @oltionchampari I think we should replace the current `useAsync`, i.e. avoid two hooks that do nearly the same thing. We can do that together next week, i.e. update the hook here, then provide PRs fixing the "build" for the apps.

puehringer:
@oltionchampari Could you refactor your usages of this hook to the new interface?
Now, instead of passing `true|false` as second parameter, you would pass the parameters of the function to be called (yours will have no parameters as the previous hook did not allow that). I.e. `true` would become `[]`, `false` would become either nothing (no 2nd parameter) or `undefined`. 
If you have any questions, just let me know. Ideally, we would be able to merge this PR and right after merge all the application PRs refactoring to this.

oltionchampari:
> @oltionchampari Could you refactor your usages of this hook to the new interface? Now, instead of passing `true|false` as second parameter, you would pass the parameters of the function to be called (yours will have no parameters as the previous hook did not allow that). I.e. `true` would become `[]`, `false` would become either nothing (no 2nd parameter) or `undefined`. If you have any questions, just let me know. Ideally, we would be able to merge this PR and right after merge all the application PRs refactoring to this.

Should i extract the actual function outside the component now 


```
  const loadOrdinoLogo = useMemo(() => async () => {
    const defaultSize = { width: 30, height: 30 };

    const plugins = PluginRegistry.getInstance().listPlugins(EP_ORDINO_LOGO);
    const plugin = plugins?.[0]; // use first registerd plugin; the order depends on import order in the phovea_registry.js of the workspace
    const module = await (await plugin.load()).factory();

    return {
      icon: module.default,
      text: plugin.text,
      width: plugin.width || defaultSize.width,
      height: plugin.height || defaultSize.height,
    };
  }, []);
```
Or can I leave the function memoized inside the component?

puehringer:
@oltionchampari For now, just refactor the second parameter. We can refine the actual functions later when this PR is merged.

thinkh (approved):
@puehringer from the Ordino side this PR can be merged. @oltionchampari has made the necessary changes and created the PRs. Let me know when you want to merge this PR so that we can merge the Ordino PRs afterwards.

PR #566, opened 2021-10-28
Data change alerting

dvtschachinger:
alert

This is how far I got by now. What I should fix next is the bugging accordion (which one should be open when) and that alerts are not "run" automatically when created. 
Also removed the hard codings from the range filter.

There are still some lint errors I didn't know how to correct properly (build error).


dvtschachinger:
I know solved all the problems except the lint errors for the circleci builds.
Also improved the deep_diff output,  improved the UI a bit and it now visualizes not only new ids but also details of the new objects, removed once and changed data.

puehringer:
@dvtschachinger Regarding the linting:
You can change the `lint` command in the package.json to `tslint --fix ...` to autofix all fixable issues. You shouldn't check it in though because then the linting happens during CI (i.e. nothing happens). 
The variable names (react components exported as const) can be temporalily disabled via 
```
// tslint:disable-next-line:variable-name
```

For the backend, you can always `docker-compose exec api bash`, then `cd tdp_core` and then `flake8` to test the linting live.

PR #570, opened 2021-11-05
Change default `font-family` to Roboto

dvdanielamoitzi:
* removed Monda and Yantramanav
* added Roboto as our new font-family


thinkh (commented):
I found some issues in a really quick test from my side.

**Wrapped tabs in LineUp side panel**

grafik

**Font size in LineUp**

The font size in LineUp is now too large again and should be decreased so that the default column can stay the same and the `g` is not truncated.

grafik

thinkh (approved):
I decreased the font size for the rankings and side panel to `13.4px` which corresponds with the visual font size of the previous font Yantramanav.

**Yantramanav**
ordino-daily caleydoapp org_app_ (4)

**Roboto**
localhost_8080_app_


**Diff**
grafik

The diff shows that the rows are almost the same. The side panel is highlighted because the width was increased to `24 em` to provide more space for the tabs. It is also appearent that the font size of the app header and detail view chooser increased with the new font. 

@dvdanielamoitzi Please check again if the changes are fine from the UI perspective. Especially, if the difference between the ranking side panel and the detail view chooser is ok for you?

@puehringer Please test the change again in other apps.

PR #578, opened 2021-11-19
Added clone() to Permission

puehringer:
**Summary**

Enabled the buddies flag for permission de/encoding.

Added clone() function to Permission. Added hints that many functions are ignoring the buddies field (i.e. we should remove those and assume buddies to be included by default --> it has been years...)

Also added a new `.tdp-busy-partial-overlay` class which is the same as `.tdp-busy-overlay` but does **NOT** hide the children, only overlays it with light opacity.


thinkh (approved):
I've tested this PR with Ordino Public. I've added some breakpoints to the changed lines, and tested the permission form of sessions, namedsets and uploaded files. But in non of the cases the debugger was stopping at the breakpoints. So I assume that your PR is fine and does not break with Ordino.

PR #583, opened 2021-11-24
Fix provenance replay

dvdanielamoitzi:
wait for the obj ref to be defined when replaying the provenance graph


puehringer (approved):
Works for me, for you as well @thinkh?

thinkh (approved):
Works as before in my tests with Ordino.

PR #592, opened 2021-12-02
Fixed tour backdrop blocker

puehringer:
The backdrop blocker of tours was just a 0x0 div, not blocking any action.


thinkh (approved):
Looks good to me.

PR #593, opened 2021-12-02
Unified the two login menus into single one

puehringer:
**Summary**
Merges the two login menus into a single one. The previous login menu just added a single prop: https://github.com/datavisyn/tdp_core/blob/b6dd2438dff4b9ec5a9e2e3a404e4a9b49351213/src/menu/LoginMenu.ts

**Breaking changes**
Generally, everything should be compatible, but 
 - `PHOVEA_SECURITY_FLASK_LoginMenu` is now called `LoginMenu`
 - `PHOVEA_SECURITY_FLASK_ILoginMenuOptions` is now called `ILoginMenuOptions`
 - `ILoginMenuAdapter` was removed because it is actually the AppHeader...


thinkh (approved):
I tested this PR with Ordino Public and did not notice any changes.

PR #594, opened 2021-12-06
Add compressor to remove consecutive nodes

thinkh:
Fixes https://github.com/Caleydo/ordino/issues/350

### Summary

* Add test and documentation for `Compressor.lastOnly()`
* Add new `Compressor.removeConsecutiveNodes()`


thinkh:
@puehringer We discovered and fixed this issue in Ordino for selections. I saw that `lastOnly` is also used for TDP selections, should we change this usage, too?

https://github.com/datavisyn/tdp_core/blob/afa673c5fb6ebf389e9f01c8661bf9d654b5540f/src/base/Selection.ts#L87-L89

dvvanessastoiber (approved):
Works in my tests, thank you for the changes! :+1:

puehringer:
@thinkh I think we should change the tdp-selections as well, maybe also in this PR for unified testing?

PR #600, opened 2021-12-16
Added async value resolver to vis components

puehringer:


thinkh:
@puehringer what is missing to merge this PR?

puehringer:
@thinkh This PR just adds the interface and PCP-changes. There is still the refactoring missing for the other vis components, but @dvzacharycutler is working on it (?).

dvzacharycutler:
All of the Vis components are now async, with a caleydo loading icon while waiting. 
async_loading

@puehringer I changed the useAsync function to set the value/error before setting a status, and wanted to make sure this was not intentional for you to have them in the opposite order.

package.json
```diff
@@ -93,6 +93,7 @@
     "md5": "^2.3.0",
     "papaparse": "~5.3.0",
     "phovea_clue": "github:phovea/phovea_clue#develop",
+    "plotly": "^1.0.6",
```
thinkh, line 96: Why do we need the "[Plotly Node API](https://www.npmjs.com/package/plotly)" as dependency here?

```suggestion
```

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -123,25 +75,61 @@ export class GeneralVisWrapper extends EventHandler {
...
-                    id: c.label + (<any> c)._id
-                },
-                values: data.map((d, i) => {
-                    return {id: d._id, val: d[(<any> c).column] ? d[(<any> c).column] : c.type === 'number' ? null : '--'};
-                }),
-                type: c.type === 'number' ? EColumnTypes.NUMERICAL : EColumnTypes.CATEGORICAL
+        const getColumnInfo = (column: Column): ColumnInfo => {
+            return {
+                name: column.getMetaData().label.replace(/(<([^>]+)>)/gi, ''),
+                description: column.getMetaData().summary.replace(/(<([^>]+)>)/gi, ''),
```
thinkh, line 89: Could you please add a short human-readable comment what the regex does? that makes it easier to understand. Thanks.

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -123,25 +75,61 @@ export class GeneralVisWrapper extends EventHandler {
...
+                // TODO: What kind of id to use?
+                id: column.fqid,
+            };
+        };
+
+
+        // wait for 2 seconds
```
thinkh, line 96: Are we still waiting here or can that be removed?

```suggestion
```

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -123,25 +75,61 @@ export class GeneralVisWrapper extends EventHandler {
...
+        const getColumnValue = async <T,>(column: ValueColumn<T>) => {
+
+            if (column.isLoaded()) {
+                return data.map((d, i) => ({id: (<IRow>d.v)._id, val: column.getValue(d)}));
+            }
+
+            return new Promise<{id: number, val: T}[]>((resolve, reject) => {
+                const timeout = setTimeout(() => {
+                    reject('Timeout');
+                }, 60000);
```
thinkh, line 107: Maybe the comment refers to this code here?

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -123,25 +75,61 @@ export class GeneralVisWrapper extends EventHandler {
...
+                cols.push({
+                    info: getColumnInfo(c),
+                    values: () => getColumnValue(c),
+                    type: EColumnTypes.NUMERICAL
+                });
+            }
+            if(c instanceof CategoricalColumn) {
+                cols.push({
+                    info: getColumnInfo(c),
+                    values: () => getColumnValue(c).then((res) => res.map((v) => v.val ? v : {...v, val: '--'})),
```
thinkh, line 127: Is `val: '--'` supposed to be a missing value? If yes, would `null` or `undefined` not work? [LineUp should replace any of these with a missing value dash](https://github.com/lineupjs/lineupjs/blob/9759adcdb78f857738e13849431341a835e20d2b/src/model/missing.ts#L3-L26).

src/lineup/internal/GeneralVisWrapper.ts
```diff
@@ -123,25 +75,61 @@ export class GeneralVisWrapper extends EventHandler {
...
+                    // TODO: This is required?
+                    colors: null,
```
thinkh, line 129: If no value is ever assigned to `colors` it could be removed. Or what does this TODO mean?

src/vis/layoutUtils.ts
```diff
@@ -132,3 +132,12 @@ export function beautifyLayout(traces: PlotlyInfo, layout: any) {
...
     return layout;
 }
+
+export function resolveColumnValues(columns: VisColumn[]) {
+    return Promise.all(columns.map(async (col) => ({...col, resolvedValues: await col.values()})));
+}
+
+export async function resolveSingleColumn(column: VisColumn) {
+    if(!column) {return null;}
+    return ({...column, resolvedValues: await column.values()});
```
thinkh, line 142:
```suggestion
    if(!column) {
        return null;
    }
    return {
        ...column,
        resolvedValues: await column.values()
    };
```

src/vis/pcp/utils.ts
```diff
@@ -20,12 +22,13 @@ const defaultConfig: IPCPConfig = {
...
 
 export function pcpMergeDefaultConfig(
-    columns: (NumericalColumn | CategoricalColumn)[],
+    columns: VisColumn[],
     config: IPCPConfig,
 ): IVisConfig {
     const merged = merge({}, defaultConfig, config);
 
     if(merged.numColumnsSelected.length === 0 && columns.length > 1) {
+        // TODO: Bug. It is always selecting the last two columns, no matter their type.
```
thinkh, line 31: If this is a bug, please open an issue and provide some information how to reproduce the error, the observed and expected behavior. Afterwards you can also add the issue link here, so that (other) developers have some context for this bug. The issue also helps to schedule a possible fix. Thanks.

src/vis/bar/utils.ts
```diff
@@ -142,14 +142,14 @@ function setPlotsWithGroupsAndMultiples(columns, config, plots, scales, plotCoun
...
-                    y: !vertFlag ? [...new Set(catCurr.values.map((v) => v.val))] : groupedLength,
+                    x: vertFlag ? [...new Set(catCurr.resolvedValues.map((v) => v.val))] : groupedLength,
+                    y: !vertFlag ? [...new Set(catCurr.resolvedValues.map((v) => v.val))] : groupedLength,
                     orientation: vertFlag ? 'v' : 'h',
                     xaxis: plotCounter === 1 ? 'x' : 'x' + plotCounter,
                     yaxis: plotCounter === 1 ? 'y' : 'y' + plotCounter,
                     showlegend: plotCounter === 1 ? true : false,
                     type: 'bar',
-                    name: uniqueGroup,
+                    name: uniqueGroup as string,
```
thinkh, line 152: I assume Typescript infers `uniqueGroup` as `number|string`. I suggest to do the typecast as early as possible so that the downstream usage is easier and you do not need to typecast multiple times.

src/vis/bar/utils.ts
```diff
@@ -108,32 +109,31 @@ export function createBarTraces(
...
-    const currGroupColumn = getCol(columns, config.group) as CategoricalColumn;
-    const currMultiplesColumn = getCol(columns, config.multiples) as CategoricalColumn;
-
+    const catCurr = catColValues[0];
+    const currGroupColumn = await resolveSingleColumn(getCol(columns, config.group));
+    const currMultiplesColumn = await resolveSingleColumn(getCol(columns, config.multiples));
 
-    const uniqueGroupVals = [...new Set(currGroupColumn.values.map((v) => v.val))];
-    const uniqueMultiplesVals = [...new Set(currMultiplesColumn.values.map((v) => v.val))];
+    const uniqueGroupVals = [...new Set(currGroupColumn.resolvedValues.map((v) => v.val))];
```
thinkh, line 125: Typecast here already, which should make the typecast below obsolete.

```suggestion
    const uniqueGroupVals = [...new Set(currGroupColumn.resolvedValues.map((v) => v.val as string))];
```

src/vis/bar/utils.ts
```diff
@@ -108,32 +109,31 @@ export function createBarTraces(
...
     uniqueMultiplesVals.forEach((uniqueMultiples) => {
         uniqueGroupVals.forEach((uniqueGroup) => {
 
             const groupedLength = uniqueColVals.map((v) => {
-                const allObjs = catCurr.values.filter((c) => c.val === v).map((c) => c.id);
-                const allGroupObjs = currGroupColumn.values.filter((c) => c.val === uniqueGroup).map((c) => c.id);
-                const allMultiplesObjs = currMultiplesColumn.values.filter((c) => c.val === uniqueMultiples).map((c) => c.id);
+                const allObjs = (catCurr.resolvedValues as {id: number, val:string}[]).filter((c) => c.val === v).map((c) => c.id);
+                const allGroupObjs = (currGroupColumn.resolvedValues as {id: number, val:string}[]).filter((c) => c.val === uniqueGroup).map((c) => c.id);
+                const allMultiplesObjs = (currMultiplesColumn.resolvedValues as {id: number, val:string}[]).filter((c) => c.val === uniqueMultiples).map((c) => c.id);
```
thinkh, line 136: Do you have an existing interface somewhere that you could reuse instead of writing `{id: number, val:string}[]` multiple times?

thinkh (changes requested):
Thanks to both of you for the changes. It improves the UI a lot!

Code-wise I have some suggestions. Please take a look below.

src/lineup/internal/GeneralVisWrapper.ts
dvzacharycutler: Unfortunately lineup isnt the issue, it's plotly. I agree '--' isnt an elegant solution to this though, and am open to ideas on how these elements should be displayed in plotly.

src/vis/bar/utils.ts
dvzacharycutler: Agreed, done

src/vis/bar/utils.ts
dvzacharycutler: created one in interfaces.ts and using it here

src/vis/bar/utils.ts
dvzacharycutler: correct, and agreed, made the change that the next comment suggests

src/vis/layoutUtils.ts
dvzacharycutler: done

src/lineup/internal/GeneralVisWrapper.ts
dvzacharycutler: removed

src/lineup/internal/GeneralVisWrapper.ts
dvzacharycutler: added

package.json
dvzacharycutler: we dont, removed

src/vis/bar/utils.ts
```diff
@@ -108,32 +109,31 @@ export function createBarTraces(
...
+    const catColValues = await resolveColumnValues(catCols);
     const vertFlag = config.direction === EBarDirection.VERTICAL;
     const normalizedFlag = config.display === EBarDisplayType.NORMALIZED;
-    const catCurr = catCols[0];
-    const currGroupColumn = getCol(columns, config.group) as CategoricalColumn;
-    const currMultiplesColumn = getCol(columns, config.multiples) as CategoricalColumn;
-
+    const catCurr = catColValues[0];
+    const currGroupColumn = await resolveSingleColumn(getCol(columns, config.group));
+    const currMultiplesColumn = await resolveSingleColumn(getCol(columns, config.multiples));
```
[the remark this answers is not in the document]
dvzacharycutler: Doing it here would require casting it to something like this, when we only really care about the resolvedValues. It's possible though

`const currGroupColumn: {
    resolvedValues: {
        id: number;
        val: string | number;
    }[];
    type: EColumnTypes.NUMERICAL;
    info: ColumnInfo;
    values: ValueGetter<{
        id: number;
        val: string | number;
    }[]>;
}`

src/vis/pcp/utils.ts
dvzacharycutler: added an issue and changed to FIXME. The issue is here: https://github.com/datavisyn/reprovisyn/issues/199

src/lineup/internal/GeneralVisWrapper.ts
thinkh: Thanks for the explanation. It's fine like this (since I currently don't have a better idea). We can change it later if we want.

src/vis/bar/BarVis.tsx
```diff
@@ -113,11 +112,11 @@ export function BarVis({
         const menu = document.getElementById(`generalVisBurgerMenu${uniqueId}`);
 
         menu.addEventListener('hidden.bs.collapse', () => {
-            Plotly.Plots.resize(document.getElementById(`plotlyDiv${uniqueId}`));
+            window.dispatchEvent(new Event('resize'));
```
thinkh, line 115: Why did you change this resize event to be dispatched on the `window`? This might cause rerenders and other changes on different parts of the application. Example: Take two general vis panels in two different views/workbenches in Ordino, toggling the vis config panel will also re-render the other general vis which must not be resized. I would revert this change, unless there are other reasons that I currently don't see.

thinkh (changes requested):
Thanks for your latest changes. From my side the code is almost ready. I only want to understand why you changed the `resize` event. Once this point it solved we can merge this PR.

PR #601, opened 2021-12-16
Add `readonly` and `disabled` options to Select3

oltionchampari:
The issue is that in comments, we propagate the `readonly ` props to select3 and select3 does not do anything with it
I saw here that https://github.com/select2/select2/issues/3387#issuecomment-281481067 that readonly is not supported
by default.

This PR is a suggestion on how that could work, but two issues are still present: you can still open the select dropdown by tabbing to it and of course not all tags are visible or scrollable if the content overflows.

We could also disable the input when the readonly flag is set. That works quite nice but its ignored then in forms.

 @puehringer what do you think, do you know a better/ easier solution? Should we support this at all?
 We do not need this feature currently if it is hard to implement


src/form/elements/Select3.ts
```diff
@@ -53,6 +53,11 @@ export interface ISelect3Options<T extends Readonly<IdTextPair>> {
    * @default false
    */
   required: boolean;
+  /**
+   * readonly flag
+   * @default false
+   */
+     readonly: boolean;
```
puehringer, line 60: Indentation is slightly off. 

This PR is fine for me, but I would not invest much more time in it. We are going to refactor tdp_comments to use `react-select` at some point, which makes most of the adjustments obsolete.

puehringer (changes requested):

oltionchampari:
Added also the disabled option to select3

puehringer (approved):

PR #604, opened 2021-12-29
 Switch LineUp dependency from `next` to version number

### LINKED ISSUE #603: Switch LineUp dependency from `next` to version number
Currently, we are using the npm tag `next` for the lineupjs dependency (see https://github.com/datavisyn/tdp_core/blob/846ddb0e5f52edcc14b1acc5d4af74b483140b0e/package.json#L85). According to npm this is corresponds with version `4.3.0-alpha.0` which is older than the `latest` release (= `4.4.0`). 

grafik

I also noticed that the `next` tag hasn't been replaced with the actual version number in the last TDP Core release.

Hence, I suggest to switch to actual version numbers for lineupjs, since there is no "rolling develop release". Once a new LineUp version (`next` or `latest`) has been released we can update the dependency version number here and test our apps accordingly. Plus: we must not change the dependency from `next` to a version and back when releasing TDP Core.

thinkh:
Closes #603

### Summary

* Switch lineupjs dependency from `next` to version number or github branch (see explanation in issue #603)
* With PR https://github.com/phovea/generator-phovea/pull/439 we switched from `node-sass` to `sass` (Dart Sass) and we can now import LineUp Sass files again and also modify the Sass variables.
grafik

* Includes many new LineUp features:
  * https://github.com/lineupjs/lineupjs/pull/542: Merged soon. **Breaking change:** if you use any HTML in your column headers, you will need to use this new option instead
  * https://github.com/lineupjs/lineupjs/pull/533: Merged soon. Allows ordering of categories in categorical column.
  * https://github.com/lineupjs/lineupjs/pull/534: Merged soon.
  * https://github.com/lineupjs/lineupjs/pull/541: Allows npm install of github dep
  * https://github.com/lineupjs/lineupjs/pull/535: Selection indicator for overflowing rankings
  * https://github.com/lineupjs/lineupjs/pull/520: New violin chart visualization
  * https://github.com/lineupjs/lineupjs/pull/514: Adjust tick visualization
  * https://github.com/lineupjs/lineupjs/pull/513: Improve selection column
  * https://github.com/lineupjs/lineupjs/pull/512: **Very important and breaking change:** Allows for auto-infering data domains of columns --> for example, you now have to pass `null` or `NaN` for min/max in numerical columns, or `undefined` for categorical columns for this to take effect. This will require some changes @dvdanielamoitzi @dg-datavisyn (i.e. applications where this is done right now --> can be removed), @oltionchampari (i.e. scores, because the hacky infering is removed with this PR --> scores need to remove a proper min/max for example to infer it automatically). 
  * https://github.com/lineupjs/lineupjs/pull/479: Vis improvements


thinkh:
I still found a bug while testing. With PR lineupjs/lineupjs#480 text sanitation was introduced. The searchbox' `formatItem()` has changed. The last commit changes our format function accordingly. It also converts the `&hellip;` to the unicode pendant.

**Before**

grafik



**After**

grafik

puehringer:
@anita-steiner Because lineupjs now supports github installs again (as seen in this build), should we use github installs again or should we use actual version numbers? 
I think version numbers are to be preferred (given a fastish lineup release cycle), as a) lineup will not easily break our workspaces, and b) releases are easier as we don't need to change any versions.

oltionchampari:
Has this been reported already? Resizing the columns does not work in ordino

grafik2

Error message:
 
grafik

thinkh:
TODO @thinkh Enable `ScheduledRenderingTask` in LineUp (after evaluation in https://github.com/datavisyn/reprovisyn/issues/186).

puehringer:
@thinkh In general, the scheduled task executor works fine for me, except that the loading overlay in the columns is rather annoying. I would prefer those to only show initially, but not when "recomputing". An alternative would be a see-through overlay instead of showing only the loading indicator. 

Lineup_Async

dvdanielamoitzi (approved):

PR #612, opened 2022-01-21
General purpose vis

dvzacharycutler:
**Summary**
@dvzacharycutler TODO

**Open issues**

- [x] LineupVisWrapper:
  - [x] Why not use ```this.props.provider.getFirstRanking().on(`${Ranking.EVENT_ORDER_CHANGED}.track`, this.updateCustomVis);``` instead of the `FILTER_CHANGED`? This would 1: be more correct (as many circumstances change the order), 2: give us live-preview when using the filter histogram for example.
  - [x] Why is ```this.props.provider.on(`${LocalDataProvider.EVENT_SELECTION_CHANGED}.track`, this.updateCustomVis);``` not included as well? This would directly sync the selection without the need of doing that in `ARankingView`.
  - [x] The `updateCustomVis` can become quite expensive to compute. We should have more granular callbacks i.e. for setting selections. Also, now the `ReactDOM.render` is called with every update which can become quite expensive. Let's discuss optimization steps here @dvzacharycutler.  

- [x] For parallel coordinates, all columns should be selectable from a single select, as the distinction into two types does not make any sense for this plot. It only limits us as you cannot have Num -> Cat -> Num. 
image
- [x] Normalization does not work: 
image

- [x] Weird behaviour when filtering in/out --> Somehow the vis is not updating properly --> should be solved by the point above with reacting to `EVENT_ORDER_CHANGED`:
  - Select two points: image
  - Click Filter In (see how ranking is updating but vis is not): image
  - Click Clear Filter: (Ranking is updating and vis is updating to show 2 points only): image
  - Click Clear Filter again to show full ranking and full vis again (correct now).

- [x] Closing the sidebar does not resize the plot: 
image

- [x] `d3` is used in a few places where a "normal" `querySelector[All]` would probably be sufficient.


thinkh:
@puehringer @dvzacharycutler What is the status here and/or what is missing? What is the timeline for the merge?

PR #614, opened 2022-01-26
Ordino 2.0: Changes to allow for Visyn React views in registry

dvzacharycutler:
Small changes to allow the PluginRegistry to accept react views, called VisynViews.
Waiting for https://github.com/datavisyn/tdp_core/pull/650.


src/base/interfaces.ts
```diff
@@ -430,11 +447,29 @@ export interface IViewPlugin {
   factory(context: IViewContext, selection: ISelection, parent: HTMLElement, options?: any): IView;
 }
 
-export interface IViewWrapperDump {
-  hash: string;
-  plugin: string;
-  dumpReference: number;
-  parameters: object;
+export interface IVisynViewProps<Desc extends IVisynViewPlugin, Param extends Record<string, any>> {
```
oltionchampari, line 450: I think this should be 
```suggestion
export interface IVisynViewProps<Desc extends IVisynViewPluginDesc, Param extends Record<string, any>> {
```
Since the  `IVisynViewPluginDesc` interface should be extendable and not the `IVisynViewPlugin `
Also in my components i would have to do like `props.desc.desc` to access any values

PR #620, opened 2022-02-08
DO NOT MERGE: Ordino 2.0

dvchristianbors:
This PR is now not needed anymore, as tdp_core#develop contains the latest frontend changes of ordino-2.0 (added [here](https://github.com/datavisyn/tdp_core/pull/702)). Therefore, also tdp_core#fastapi contains those latest changes (as it contains latest tdp_core#develop). So, the FastAPI migration now became much easier, as you only have to switch tdp_core to #fastapi and it should work. 
The ordino-2.0 branch will then be archived (as it only contains the legacy JWT stuff), it will never be merged into #develop. 
We will close this PR when everything is merged and ready. For now, it is for visibility purposes only.


puehringer:
Closed in favor of the FastAPI implementation in #develop.

PR #626, opened 2022-02-15
Fixed this.itemIDType not ready yet in ARankingView

puehringer:
Fixes `this.itemIDType` not being ready when the ranking is initialized.


oltionchampari (approved):
Tested together and it works

Issue #637, opened 2022-02-16
Add Tooltips to the Filter Buttons


dg-datavisyn:
* Release number or git hash: Version: 10.0.1-SNAPSHOT
* Web browser version and OS: Google Chrome Version 98.0.4758.102 (Offizieller Build) (64-Bit) on Win 11
* Environment (local or deployed): https://ordino.dev.app.datavisyn.io/

### Steps to reproduce

1. Open a tissue data ranking
2. Open the Vis panel
3. Create a scatter chart

Tooltips would be helpful here to explain the functionality of these 3 buttons (at least the first 2)

![image.png](https://images.zenhubusercontent.com/5c3ee30c116ade5d16d9b79c/f7487a67-ef6d-4a50-b798-f74feec77e4d)

PR #660, opened 2022-03-14
Zc filter improvements

### LINKED ISSUE #636: CSS glitch in the filter buttons
* Release number or git hash: Version: 10.0.1-SNAPSHOT
* Web browser version and OS: Google Chrome Version 98.0.4758.102 (Offizieller Build) (64-Bit) on Win 11
* Environment (local or deployed): https://ordino.dev.app.datavisyn.io/

### Steps to reproduce

1. Open a tissue data ranking
2. Open the Vis panel
3. Create a scatter chart

### Expected behavior
* Maybe just write `Clear` or center the others.

dvzacharycutler:
Adding better spacing, tooltips to filter buttons. 

fix #636, #637

Should note here the filter buttons as a whole still don't really make sense, as there's no way to see if a filter has already been applied. A dedicated filter interface would probably be valuable.


src/vis/sidebar/FilterButtons.tsx
```diff
@@ -5,24 +5,47 @@ interface FilterButtonsProps {
...
+      tooltip: 'Filters any point not currently selected',
+    },
+    {
+      name: EFilterOptions.OUT,
+      tooltip: 'Filters all currently selected points',
+    },
+    {
+      name: EFilterOptions.CLEAR,
+      tooltip: 'Removes any existing filter',
+    },
```
thinkh, line 26: Please extract language strings using i18n.

I'm not sure though how you would make this work with enums, since you directly use the enum value as name. You might need to create separate language strings here.

thinkh: I also saw that there are still some language strings (especially the enum values) in other places that should of course be extracted, too. If you want to extract the language strings all together in a separate PR, please open a new issue + PR. Thanks.

thinkh (changes requested):

Issue #632, opened 2022-02-16
Avoid nonesense configurations


dg-datavisyn:
* Release number or git hash: Version: 10.0.1-SNAPSHOT
* Web browser version and OS: Google Chrome Version 98.0.4758.102 (Offizieller Build) (64-Bit) on Win 11
* Environment (local or deployed): https://ordino.dev.app.datavisyn.io/

### Steps to reproduce

1. Open a tissue data ranking
2. Open the Vis panel
3. Create a bar chart

### Observed behavior

* I can add the same attribute to categorize, group and multiples

![image.png](https://images.zenhubusercontent.com/5c3ee30c116ade5d16d9b79c/338e3801-f02c-404a-97c6-10b0a56902fc)

### Expected behavior
* If I select a column as main parameter, I should not be able to select it again as group or multiple parameter. It doesn't make any sense

PR #663, opened 2022-03-14
Zc config improvements

### LINKED ISSUE #628: Default-Scatterplot without numeric data
* Release number or git hash: Version: 10.0.1-SNAPSHOT
* Web browser version and OS: Google Chrome Version 98.0.4758.102 (Offizieller Build) (64-Bit) on Win 11
* Environment (local or deployed): https://ordino.dev.app.datavisyn.io/

### Steps to reproduce

1. Open a tissue data ranking
2. Open the Vis panel

### Observed behavior
* The vis component is launched with a scatter plot although there is no numeric data available. As a user, I can't do anything with the scatter plot.
* This also applies to the Strip Plots. They are offered as an option although I don't have any numeric data to populate the view

### Expected behavior
* Open the bar chart by default if there is no numeric data but categorical data
* We could also show the `Add Column` button there

dvzacharycutler:
Vis defaults to a bar chart if there arent 2 numerical columns. Also avoid seeing duplicate options in bar chart, such as selecting "gender" as the bar category then "gender" again for the multiples option.

fix #628, #632


src/vis/scatter/ScatterVis.tsx
```diff
@@ -64,6 +64,8 @@ export function ScatterVis({
 }) {
   const id = React.useMemo(() => uniqueId('ScatterVis'), []);
 
+  console.log(config);
+
```
thinkh, line 68:
```suggestion
```

Remove

thinkh (changes requested):
Code-wise it looks fine. Please attach a short screencast (or multiple screenshots) that documents that issue https://github.com/datavisyn/tdp_core/issues/632 has been fixed. Thanks.

thinkh (approved):
Thanks for the screencasts and your changes. Looks good to me now.

PR #666, opened 2022-03-16
Ranking component

oltionchampari:
Coming from https://github.com/datavisyn/reprovisyn/issues/204
### Summary
* Created Component Ranking that is a wrapper around LineUp. Offers the same functionality as the `ARankingView` minus the `inputSelection` and `parameter` logic. That logic is contained in `RankingViewRankingcomponent`

Some differences: The Ranking does not contain Vis stuff since we already offer that as a separate view --> How should we proceed here? There is some commenting and code clean up required eventually, but I would prefer to get some feedback before that


src/lineup/panel/PanelAddColumnButton.ts
```diff
@@ -24,18 +24,19 @@ export class PanelAddColumnButton implements IPanelButton {
...
         onClick: () => {
           this.node.classList.add('once');
           (<HTMLElement>this.search.node.querySelector('input'))!.focus();
           this.search.focus();
         },
-      }),
-    );
+      },
+    };
+    const button = new PanelButton(this.node, btnOptions);
```
oltionchampari, line 39: Copy options to avoid sharing the addColumnButtonOptionsObject between rankings

src/lineup/internal/LineUpSelectionHelper.ts
```diff
@@ -102,9 +102,9 @@ export class LineUpSelectionHelper extends EventHandler {
     }
 
     // What are these remove and add event listeners doing ? They mess with my event listeners as well
-    // this.removeEventListener();
+    this.removeEventListener();
     this.provider.setSelection(indices);
-    // this.addEventListener();
+    this.addEventListener();
```
oltionchampari, line 107: Uncomment the events, which were commented out on the base branch

src/vis/pcp/PCPVis.tsx
```diff
@@ -65,7 +76,7 @@ export function PCPVis({ config, extensions, columns, setConfig, hideSidebar = f
   }, [traces]);
 
   return (
-    <div className="d-flex flex-row w-100 h-100" style={{ minHeight: '0px' }}>
+    <div ref={plotlyDivRef} className="d-flex flex-row w-100 h-100" style={{ minHeight: '0px' }}>
       <div
         className={`position-relative d-flex justify-content-center align-items-center flex-grow-1 ${
```
oltionchampari, line 81: This fixes a bug in the Vis. This should probably be fixed in the develop branch to avoid branches diverging

src/vis/scatter/ScatterVis.tsx
```diff
@@ -80,8 +80,20 @@ export function ScatterVis({
...
+        Plotly.Plots.resize(plotDiv);
+      }
+    });
+
+    if (plotlyDivRef) {
+      ro.observe(plotlyDivRef.current);
+    }
+
     if (hideSidebar) {
       return;
```
oltionchampari, line 98: Same comment as above

PR #671, opened 2022-03-23
Allow source or target to be null in GraphEdge

puehringer:
Fixes an error when restoring a graph where either source or target is null. I honestly do not know the underlying issue why one of these might be null (there is a default value of null for both parameters). But this fixes the issue...


src/clue/graph/graph.ts
```diff
@@ -116,8 +116,12 @@ export class GraphEdge extends AttributeContainer {
...
 
   private init() {
-    this.source.outgoing.push(this);
-    this.target.incoming.push(this);
+    if (this.source) {
+      this.source.outgoing.push(this);
+    }
+    if (this.target) {
+      this.target.incoming.push(this);
+    }
```
thinkh, line 124: This is already covered in the constructor (line 106). But `init()` is also called in `restore()` in line 155. Probably the return values of `nodes()` for source and target is `null`.

thinkh (approved):
I don't know how to test this or how you produced this error, but I assume it is safe to merge.

PR #672, opened 2022-03-24
Adding filter to data view interface

dvzacharycutler:
Adding filter and filter callback


src/views/visyn/interfaces.ts
```diff
@@ -168,6 +168,17 @@ export type VisynDataViewPluginType<
      * TODO:: Type to IReprovisynServerColumn when we merge that into tdp_core
      */
     dataDesc: IServerColumn[] | any[];
+
+    /**
+     * List of items which are filtered out of the view. Ids match the idtype from 'desc.idtype'
+     */
+    idFilter: string[];
```
thinkh, line 175: For me the property name was a bit confusing. I would suggest the following renaming. What do you think?

```suggestion
    /**
     * List of items which are filtered out in this view. The ids must match the idtype from 'desc.idtype'
     */
    filteredOutIds: string[];
```

src/views/visyn/interfaces.ts
```diff
@@ -168,6 +168,17 @@ export type VisynDataViewPluginType<
...
+
+    /**
+     * Callback when the Filter changed.
+     * @param selection New Filter.
+     */
+    onIdFilterChanged(idFilter: React.SetStateAction<string[]>): void;
```
thinkh, line 181: Same here:

```suggestion
    /**
     * Callback when ids that are filtered out are changed.
     * @param filteredOutIds New Filter.
     */
    onFilteredOutIds Changed(filteredOutIds: React.SetStateAction<string[]>): void;
```

PR #674, opened 2022-03-28
Density hex vis

dvzacharycutler:
Adding initial version of hexbin visualization, using d3 and react. Math behind hexbins, panning, selection, etc is d3, while all rendering logic is react.

Intended to be a 1.0, and there are still many features to be added and likely bugs to be fixed. 

Additional to the hexbins, this PR also includes a move from all imports of "d3" in tdp_core to a pinned "d3v3" import. Additionally, all imports in tdp_core which were using package specific imports based on d3v4 imports are now using a pinned "d3v4" import. 

Currently, this causes one known problem which is the breaking of pan/zoom and selection of the hexbin plots, because the events are using a different version of d3. Open to solutions for this! Screen casts are from removing all parts of d3 which cause problems, such as the old imports, which is of course not possible on develop.

https://user-images.githubusercontent.com/87644761/161103866-4faac2c2-3321-46d3-9306-dc8f134598c1.mp4


src/vis/density/DensityVis.tsx
```diff
@@ -0,0 +1,102 @@
+import * as React from 'react';
+import { merge, uniqueId } from 'lodash';
+import { useMemo } from 'react';
+import { VisColumn, IVisConfig, IDensityConfig, EScatterSelectSettings } from '../interfaces';
+import { DensityVisSidebar } from './DensityVisSidebar';
+// eslint-disable-next-line import/no-cycle
```
thinkh, line 6:
```suggestion
```
Can we remove this line? Locally my ESLint was not complaining when I removed it.

src/vis/density/DensityVis.tsx
```diff
@@ -0,0 +1,102 @@
...
+                      <HexagonalBin
+                        selectionCallback={selectionCallback}
+                        selected={selected}
+                        config={config}
+                        columns={[columns.find((col) => col.info.id === xCol.id), columns.find((col) => col.info.id === yCol.id)]}
+                      />
+                    );
+                  }
+
+                  return <div key={`${xCol.id}hist`}>hello world</div>;
```
thinkh, line 68: Can we replace `hello world` with something meaningful?

src/vis/density/PieChart.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }, [radius]);
+
+  const id = React.useMemo(() => uniqueId('PieNum'), []);
+
+  return (
+    <g style={{ transform }}>
+      {pie(data).map((slice, i) => {
+        // TODO: Why are indexes bad in the key? how else to do this? Also, I think the typings for arc are wrong, which is why im typing slice to any
+        // eslint-disable-next-line react/no-array-index-key
+        return <path key={`${id}, ${i}`} d={arc(slice as any)} style={{ fill: colorScale ? colorScale(dataCategories[i]) : 'cornflowerblue' }} />;
```
thinkh, line 30: Maybe you can apply a similar solution as proposed in [this article](https://medium.com/stationfive/how-to-create-a-pie-chart-with-d3-js-and-react-hooks-part-1-81bcd7f39b32) (jump to _The simplest way using SVG elements_)?

Additionally, locally my ESLint was not complaining when I removed the eslint-disable line.

src/vis/density/utils.tsx
```diff
@@ -0,0 +1,107 @@
...
+    .map((c) => columns.find((col) => col.info.id === c.id) as VisNumericalColumn);
+
+  const numColVals = await resolveColumnValues(numCols);
+
+  const colorColVals = await resolveSingleColumn(colorColumn ? columns.find((col) => col.info.id === colorColumn.id) : null);
+
+  return { numColVals, colorColVals };
+}
+
+export function cutHex(path: string, radius: number, start: number, sixths: number): string {
```
thinkh, line 76: Please add a function documentation and describe what this function does. Currently, it's not very obvious from the function name + parameters.

src/vis/density/utils.tsx
```diff
@@ -0,0 +1,107 @@
...
+const defaultConfig: IDensityConfig = {
+  type: ESupportedPlotlyVis.DENSITY,
+  numColumnsSelected: [],
+  color: null,
+  isOpacityScale: true,
+  isSizeScale: false,
+  hexRadius: 16,
+  dragMode: EScatterSelectSettings.RECTANGLE,
+  hexbinOptions: EHexbinOptions.COLOR,
+};
```
thinkh, line 30: This seems to be equal to the code in Vis.tsx. Maybe export it here and use it in Vis.tsx?

package.json
```diff
@@ -85,20 +85,14 @@
...
-    "d3-polygon": "^1.0.3",
-    "d3-quadtree": "^1.0.3",
-    "d3-scale": "^2.1.0",
-    "d3-selection": "^1.3.0",
-    "d3-shape": "^1.2.0",
-    "d3-zoom": "^1.7.1",
+    "d3-hexbin": "^0.2.2",
+    "d3v3": "npm:d3@~3.5.17",
+    "d3v4": "npm:d3@~4.13.0",
+    "d3v7": "npm:d3@~7.3.0",
```
thinkh, line 95: @puehringer Do you agree with the new D3 imports/aliases? Zach merged the former separate d3v4 imports into a single `d3v4` import. Are we then still fine with the bundle size?

src/vis/density/HexagonalBin.tsx
```diff
@@ -0,0 +1,411 @@
...
+          yScale(currentY.filteredValues[i].val as number),
+          currentColorColumn ? currentColorColumn.filteredValues[i].val : '',
+          c.id,
+        ]);
+      });
+    }
+
+    // TODO: Im cheating a bit here by appending the id/color value to each hex, breaking the types.
+    // is there a better way to type this?
+    return d3Hexbin(inputForHexbin) as unknown as HexbinBin<[number, number, string, string]>[];
```
thinkh, line 213: As far as I understand, the original HexbinBin typing only supports `[number, number]` and you added color and id as additional information. Where are you using/reading the two additional properties? It doesn't seem to be consumed by the `.hexbin()` function above.

src/vis/density/HexagonalBin.tsx
```diff
@@ -0,0 +1,411 @@
...
+        d3v7.select(this).call(brush.move, null);
+      }),
+    );
+  }, [width, height, id, hexes, selectionCallback, config.dragMode, xZoomTransform, yZoomTransform, zoomScale, xScale, yScale]);
+
+  // TODO: svg elements seem weird with style/classNames. I can directly put on a transform to a g, for example, but it seems to work
+  // differently than if i use style to do so
+  return (
+    <div ref={ref} className="mw-100">
+      <svg id={id} style={{ width: width + margin.left + margin.right, height: height + margin.top + margin.bottom }}>
```
thinkh, line 361: Setting the width and height of an SVG element is fine. Do you need to add the `px` unit for each of them if you use the `style`? Otherwise you could also use the `width` and `height` attribute, since they don't require a unit.

src/vis/density/HexagonalBin.tsx
```diff
@@ -0,0 +1,411 @@
...
+    zoom.on('zoom', (event: D3ZoomEvent<any, any>) => {
+      const { transform } = event;
+
+      const newX = transform.rescaleX(xScale);
+      const newY = transform.rescaleY(yScale);
+
+      // Question: I dont think this should be a ref, because it doesnt actually cause a re render. Only the other setters below make it work, if i moved them above this code there would be bugs.
+      // But when I made it a useState object it didnt work with the object.
+      xZoomedScale.current = newX;
+      yZoomedScale.current = newY;
```
thinkh, line 296: It's hard for me to figure this out from the code. I think a direct conversation about this point would make sense.

package.json
puehringer: I think the bundle size is a neglible caveat of this solution, but we should still move forward, as this would ideally give us a way of removing d3[v3|v4] in a more structured fashion. Also, the d3v4 dependency will not be there for long, as it is coming from tdp_scatterplot, something that is deprecated and will be removed soonish.

dvzacharycutler:
> 

Regarding the placement of the toolbar this was intentional, for no real specific reason. I can move it to the center no problem. 
For the selection problems, that's a side effect of the d3 import problem I mentioned in the initial description. There should be a different encoding for selected hexbins, as is shown in the video, but right now selection wont work at all, and I've wasted more time than I care to admit trying to fix it. Though selection being cleared when switching between scatterplot -> hexbin is something that I think is storybook specific right now.

src/vis/density/utils.tsx
dvzacharycutler: Done

src/vis/Vis.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import d3 from 'd3';
+import d3 from 'd3v3';
```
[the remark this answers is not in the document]
dvzacharycutler: Updated all of the imports in vis to d3v7, and made sure the imports are all "import * as d3v from 'd3v7'" instead of the above example.

src/vis/density/DensityVis.tsx
dvzacharycutler: Just removed it for now. This is what is rendered in the diagonal of the small multiples. It should probably be a histogram like the scatterplot has it, but I will leave that to a future ticket

src/vis/density/HexagonalBin.tsx
dvzacharycutler: I'm using them in the SingleHex component to assign categories/figure out if a hex is selected. D3 itself does not use those values at all, but it also does not remove them making them useful for me. Otherwise I'd need to keep a separate variable which is up to date with the d3 hexes that has this information.

src/vis/Vis.tsx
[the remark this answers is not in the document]
thinkh: Thanks. 👍

src/vis/density/DensityVis.tsx
thinkh: Yes, the histogram would probably the preferred visualization in this case. Could you create a new issue then we can see if and when we want to schedule it. Thanks.

package.json
thinkh: Ok, sounds good.

src/vis/density/HexagonalBin.tsx
thinkh: I see. I'm fine with this solution. The only thing I would change then is to remove the "TODO" + question. Then just document this addition in a comment and mention where it is currently used. This way future developer can easier follow-up and navigate in the code. Thanks.

dvzacharycutler:
@puehringer We discovered the d3 problems were mainly from the separate d3v4. I converted all of the old scatterplot imports to d3v7, and the types dont seem to get mad and it compiles, but I have no real way of testing this at runtime. Any suggestion on how to test the Scatterplot?

puehringer:
@dvzacharycutler Thomas can test the AScatterplot using this branch in one of the dTiles applications, as there are some views using it. @dvtschachinger could you verify everything works for the AScatterplot views when using this branch? Maybe sync with @dvzacharycutler and test it together.

dvtschachinger:
@puehringer, @dvzacharycutler tried out the branch in Target360. All views using the AScatterplot work without errors and brushing works when available.

thinkh (approved):
@dvtschachinger Thanks for testing the scatterplot component and your feedback.

@dvzacharycutler Thanks for your bug fixes and changes. Code-wise it looks good to me. 👍

Issue #675, opened 2022-03-29
Filter Dropdown in Detail View does not show any options


dvvanessastoiber:
* Release number or git hash: fd8bf51d5651688d8542a3b877c47b6be4dc2244
* Web browser version and OS: Chrome 99
* Environment (local or deployed): both

### Steps to reproduce

1. Open https://ordino-daily.caleydoapp.org/#/
2. Open the `Cancer Gene Census` dataset
3. Select two items
4. Open a detail view, e.g. Expression
5. Open the Filter Select

### Observed behavior
* The Filter select (FormMap) does not show any options, as shown in the screenshot below
![ordino](https://user-images.githubusercontent.com/52395160/160563628-b1cde651-7ab0-4718-a930-302e2925a306.png)

### Expected behavior
* The Filter shows a list of options to filter the dataset

thinkh:
@dvvanessastoiber Is this an Ordino issue that the FormMap component does not receive data or is it a bug in the FormMap code? Could you please check if there are errors on the dev console?

dvvanessastoiber:
> @dvvanessastoiber Is this an Ordino issue that the FormMap component does not receive data or is it a bug in the FormMap code? Could you please check if there are errors on the dev console?

I do not get any errors related to that in the console. Could it be the case that only the formMap component does not receive the data?

thinkh:
Thanks. You could also cross-check with Ordino Public and/or the local workspace.

dvvanessastoiber:
> Thanks. You could also cross-check with Ordino Public and/or the local workspace.

I can reproduce this in both local workspaces (Ordino public and BI)

dvvanessastoiber:
@thinkh and I have discovered the following issues:

- `ResolveNow.resolveImmediately()` has been replaced with `Promise.resolve()`
- The Filter FormMap tries to use the options of the FormSelect, but the options are not loaded / the dropdown items are not ready at this point
- Crucial points:
  - Waiting for [dependents in FormSelect](https://github.com/datavisyn/tdp_core/blob/develop/src/form/elements/FormSelect.ts#L113-L135)
  - `EVENT_INITIAL_VALUE` event in [FormMap](https://github.com/datavisyn/tdp_core/blob/develop/src/form/elements/FormMap.ts#L239-L241)

PR #676, opened 2022-03-30
Switched FormSelect to use ResolveNow to mitigate race condition

puehringer:
Closes https://github.com/datavisyn/tdp_core/issues/675


dvvanessastoiber (approved):
Works in my tests, thank you for the implementation! :+1:

PR #683, opened 2022-04-01
Added enhancements of JWT security store by Markus

puehringer:


tdp_core/security/manager/jwt_security_manager.py
```diff
@@ -212,69 +106,117 @@ def logout(self):
...
+    # Append the access token as cookie
+    set_access_cookies(response, access_token)
+    return response
+
+
+@login_required()
+@blp.route('/logout', methods=['GET', 'POST'])
+@no_cache
+def logout():
+    verify_jwt_in_request()
```
puehringer, line 199: As we already do this in the `current_user` property (triggered by the next call), can we remove it here?

tdp_core/security/manager/jwt_security_manager.py
```diff
@@ -212,69 +106,117 @@ def logout(self):
...
+            _log.exception(f'Error setting cookie {cookie} to logout response')
+    # Also unset the access token cookies
+    unset_jwt_cookies(response)
+    return response
+
+
+@blp.route('/loggedinas', methods=['GET', 'POST'])
+@no_cache
+def loggedinas():
+    verify_jwt_in_request()
```
puehringer, line 217: As we already do this in the `current_user` property (triggered by the next call), can we remove it here?

tdp_core/security/manager/jwt_security_manager.py
```diff
@@ -100,105 +79,20 @@ def refresh_expiring_jwts(response):
...
     @property
-    def current_user(self) -> PhoveaServerUser:
-        try:
-            return self._load_user_from_request(request)
-        except Exception:
-            _log.exception('Error fetching user from current request')
-            return None
+    def current_user(self) -> Optional[PhoveaServerUser]:
+        with self.app.app_context():  # used for jwt request
+            verify_jwt_in_request()
```
puehringer, line 94: If I remember correctly this methd would throw as soon as no token is present. As we want to allow loading users via the other functions as well (like `_api_key_from_headers` or essentially `_delegate_stores_until_not_none`), we should either add [optional=True](https://flask-jwt-extended.readthedocs.io/en/stable/api/#flask_jwt_extended.verify_jwt_in_request) or wrap it in an exception handler, ultimately triggering the other functions.

puehringer: I would prefer the exception handler I think, because an invalid token should not abort if the other methods return a valid user.

tdp_core/security/manager/jwt_security_manager.py
```diff
@@ -1,70 +1,38 @@
...
+from datetime import datetime, timedelta, timezone
 from functools import wraps
-from typing import Dict, List, Optional, Union
-from flask import request, abort, jsonify, render_template_string, Flask, Request
-from phovea_server.security import SecurityManager as PhoveaServerSecurityManager, User as PhoveaServerUser
-from phovea_server.plugin import list as list_plugins
-from phovea_server.ns import no_cache
+from typing import Dict, Optional
+
+from flask import request, abort, jsonify, render_template_string, Flask, Request, Blueprint, current_app, _request_ctx_stack
```
puehringer, line 8: Some flake8 issues remain due to imports: https://app.circleci.com/pipelines/github/datavisyn/tdp_core/1413/workflows/ec003492-643c-41bc-b3df-0945f273751f/jobs/4133

puehringer (commented):
@dvmarkusvogl Looks good me! Some minor things which we might want to address first. Will do a functional test afterwards.

PR #687, opened 2022-04-04
Different aggregation types in bar chart

dvzacharycutler:
Adding different aggregation types in bar charts. Creating a new UI dropdown for aggregation type, followed by a aggregation column selector for all of the aggregation types except for count. Also disabled those aggregate types if there arent any numerical columns, with an info icon explaining why. 

Also has some changes to ensure that the vis component is never rendering an inconsistent visConfig object, which it previously was for one render after changing types. 

Also one small change in useAsync to ensure useAsync doesnt throw errors if the component gets unmounted in the middle of the async call

https://user-images.githubusercontent.com/87644761/161645338-ef011aa9-69e1-4104-95c3-9646c1fb01ea.mp4


src/vis/bar/utils.ts
```diff
@@ -31,30 +35,118 @@ const defaultConfig: IBarConfig = {
...
+        (curr) => median((aggColValues as VisNumericalValue[]).filter((c) => categoricalMap[c.id] === curr && !Number.isNaN(c.val)).map((c) => c.val)) || 0,
+      ),
+    );
+  }
+
+  return sum(
+    categoricalOptions.map(
+      (curr) => max((aggColValues as VisNumericalValue[]).filter((c) => categoricalMap[c.id] === curr && !Number.isNaN(c.val)).map((c) => c.val)) || 0,
+    ),
+  );
```
thinkh, line 133: Please refactor accordingly.

src/vis/bar/utils.ts
```diff
@@ -31,30 +35,118 @@ const defaultConfig: IBarConfig = {
...
+}
+
 async function setPlotsWithGroupsAndMultiples(
   columns: VisColumn[],
   catCol: VisCategoricalColumn,
+  aggType: EAggregateTypes,
+  aggregateColumn: VisNumericalColumn,
```
thinkh, line 140:
```suggestion
  aggregateType: EAggregateTypes,
  aggregateColumn: VisNumericalColumn,
```

Why one is abbreviated and one not?

src/vis/bar/utils.ts
```diff
@@ -195,6 +326,8 @@ async function setPlotsWithMultiples(
 
 async function setPlotsBasic(
   columns: VisColumn[],
+  aggType: EAggregateTypes,
+  aggregateColumn: VisNumericalColumn | null,
```
thinkh, line 330:
```suggestion
  aggregateType: EAggregateTypes,
  aggregateColumn: VisNumericalColumn | null,
```

src/vis/bar/utils.ts
```diff
@@ -151,12 +269,16 @@ async function setPlotsWithGroups(
 async function setPlotsWithMultiples(
   columns: VisColumn[],
   catCol: VisCategoricalColumn,
+  aggType: EAggregateTypes,
+  aggColumn: VisNumericalColumn,
```
thinkh, line 273:
```suggestion
  aggregateType: EAggregateTypes,
  aggregateColumn: VisNumericalColumn,
```

src/vis/bar/utils.ts
```diff
@@ -104,12 +208,16 @@ async function setPlotsWithGroupsAndMultiples(
 async function setPlotsWithGroups(
   columns: VisColumn[],
   catCol: VisCategoricalColumn,
+  aggType: EAggregateTypes,
+  aggColumn: VisNumericalColumn,
```
thinkh, line 212:
```suggestion
  aggregateType: EAggregateTypes,
  aggregateColumn: VisNumericalColumn,
```

Name equal parameters equal also across functions.

src/vis/Vis.tsx
```diff
@@ -69,54 +74,85 @@ export function Vis({
   externalConfig?: IVisConfig;
   hideSidebar?: boolean;
 }) {
-  const [visConfig, setVisConfig] = React.useState<IVisConfig>(
-    externalConfig || columns.filter((c) => c.type === EColumnTypes.NUMERICAL).length > 1
+  // Each time you switch between vis config types, there is one render where the config is inconsistent with the type before the merge functions in the useEffect below can be called.
+  // To ensure that we never render an incosistent config, keep a consistent and a current in the config. Always render the consistent.
+  // TODO:: probably change the names, especially "current" and "inconsistentVisConfig" to something more useful.
```
thinkh, line 79: From the state properties, I would either expect something like `consistent` + `inconsistent` or `current` + `previous`. If they are matching you could also name your local variables `visConfig` and `inconsistentVisConfig`.

In general, however, we should try to solve the root of the inconsistent vis config. @puehringer do you have an idea if there is still some issue with our vis config setup?

thinkh (changes requested):
Thanks for this nice feature. Please see my comments below.

src/vis/bar/utils.ts
```diff
@@ -31,30 +35,118 @@ const defaultConfig: IBarConfig = {
...
+  }
+  if (aggType === EAggregateTypes.MED) {
+    return categoricalOptions.map((curr) =>
+      median((aggColValues as VisNumericalValue[]).filter((c) => categoricalMap[c.id] === curr && !Number.isNaN(c.val)).map((c) => c.val)),
+    );
+  }
+
+  return categoricalOptions.map((curr) =>
+    max((aggColValues as VisNumericalValue[]).filter((c) => categoricalMap[c.id] === curr && !Number.isNaN(c.val)).map((c) => c.val)),
+  );
```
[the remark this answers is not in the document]
dvzacharycutler: This is great!

src/vis/Vis.tsx
thinkh: We agreed in a chat that we will keep the solution for now as it is and might refactor it later with issue https://github.com/datavisyn/reprovisyn/issues/304.

thinkh (approved):

PR #692, opened 2022-04-14
External Config Updates

dvzacharycutler:
fix https://github.com/datavisyn/reprovisyn/issues/304

Added an external config callback for the Vis component, which is called whenever the stable version of the vis config is changed. Also removed all internal state from the VisSidebar component, making it a fully controlled component based on the incoming config.

Open to suggestions on this code https://github.com/datavisyn/tdp_core/pull/692/files#diff-3eb51bfc3f050754fa611f761b78079520549ca75cc2f70ee39fd17c43af019dR119-R124. Currently if you remove the JSON.stringify this creates an infinite loop. I understand why this happens, and am actually more confused as to how this is not currently a problem, as this code is identical to what is in the current develop VisSidebar file, here https://github.com/datavisyn/tdp_core/blob/657d51be90b3c23b3f777b73a9b25f8fd8c40dab/src/vis/VisSidebar.tsx#L41-L44


src/vis/Vis.tsx
```diff
@@ -113,6 +116,12 @@ export function Vis({
         },
   );
 
+  const setExternalConfigRef = useSyncedRef(setExternalConfig);
+  useEffect(() => {
+    setExternalConfigRef.current?.(visConfig);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [JSON.stringify(visConfig), setExternalConfigRef]);
```
thinkh, line 123: @puehringer Can you please have a look why `JSON.stringify(visConfig)` is necessary here? Removing this will cause an infinite loop, according to Zach's initial post above.

thinkh (approved):
Thanks for this refactoring. In general the changes look fine to me, but I don't know either why a `JSON.stringify()` is necessary. Maybe @puehringer can have a look.

src/vis/Vis.tsx
puehringer: @thinkh Well the `JSON.stringify()` makes it comparable while ignoring references, i.e. `{} !== {}` but `'{}' === '{}'`. Therefore, somewhere in your code you are not changing the "value"/string representation, but still assign a new object to the visConfig. You could also do a deepEqual check *before* using the setter to avoid a rerender completely.

dvzacharycutler: To follow up on this after talking to Michael, the problem here was not actually in the vis component, but an issue of not memoizing the props being passed to the vis component in ordino 2.0. Changes were made to the https://github.com/Caleydo/ordino/pull/378 PR to account for this. The stringify was also removed.

thinkh (approved):
Thanks for refactoring/fixing the `JSON.stringify()` code. We can merge this PR once the FastAPI PRs are all merged.

src/vis/Vis.tsx
thinkh: Thanks for the explanation.

thinkh:
@dvzacharycutler Please update this branch one more time and update the merge conflicts. Thanks.

PR #700, opened 2022-04-21
get_nested shouldn't return None

dvmarkusvogl:
See code. It's 2 lines.
Breaks alb_security_store check enabled as i don't have any alb_security_store check.

General code hint: `a if a else b` == `a or b`

### DIFF
--- a/tdp_core/settings/model.py
+++ b/tdp_core/settings/model.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Dict, List, Literal, Union
+from typing import Any, Dict, List, Literal, Optional
 
 from pydantic import BaseModel, BaseSettings, Extra, Field
 
@@ -111,14 +111,14 @@ class GlobalSettings(BaseSettings):
     def is_development_mode(self) -> bool:
         return self.env.startswith("dev")
 
-    def get_nested(self, key: str, default: Any = None) -> Union[Any, None]:
+    def get_nested(self, key: str, default: Optional[Dict] = None) -> Dict:
         # TODO: Set deprecated
         keys = key.split(".")
         plugin_id = keys[0]
         dic = self.dict(include={plugin_id})
         for key in keys:
             dic = dic.get(key, None) if dic else None
-        return dic if dic is not None else default
+        return dic or default or {}
 
     class Config:
         extra = Extra.allow


tdp_core/settings/model.py
```diff
@@ -111,14 +111,14 @@ class GlobalSettings(BaseSettings):
     def is_development_mode(self) -> bool:
         return self.env.startswith("dev")
 
-    def get_nested(self, key: str, default: Any = None) -> Union[Any, None]:
+    def get_nested(self, key: str, default: Optional[Dict] = None) -> Dict:
```
puehringer, line 114: How come the default value should be `Dict`? What if I want the leaf value which could be anything from `str` to `int` to `bool` etc.?

tdp_core/settings/model.py
```diff
@@ -111,14 +111,14 @@ class GlobalSettings(BaseSettings):
...
         # TODO: Set deprecated
         keys = key.split(".")
         plugin_id = keys[0]
         dic = self.dict(include={plugin_id})
         for key in keys:
             dic = dic.get(key, None) if dic else None
-        return dic if dic is not None else default
+        return dic or default or {}
```
puehringer, line 121: `dic if dic is not None else default` is a very different to `dic or ...`, as I explicitly want to allow falsy values like 0 or `False`. 
I think you somehow expect this function to always return a dict, or am I missing something? This function can return anything really, as long as it is a JSON compatible datatype.

puehringer (commented):
I am not sure if I understood your changes, as they break the function in how it is intended to be used. I think you somehow expect this function to always return a dict, or am I missing something? This function can return anything really, as long as it is a JSON compatible datatype.

puehringer:
@dvmarkusvogl Your hint is correct, i.e. `a if a else b` == `a or b`, however the check in the code is `a if a is not None else b`, which is very different and != `a or b`. Just use `a = 0` or `a = []`, both very much valid return values.

tdp_core/settings/model.py
dvmarkusvogl: Every usage I've encountered is getting a "sub config" (dict), which is also what I'd assume a "settings.get_nested()" would do. Get some settingS, plural, of a nested kind.

And I've encountered usage (in this codebase) that looks sane when observed
```
_conf = manager.settings.get_nested("tdp_core.security.store.
...
if _conf.get("enable", False):
```
But explodes in your face, because the _conf is None.

And in general, the only code where it makes a difference is where dic would be a false-ish value (False, 0), it would return a {} in the end, which is treated as False by any `if/not` anyways.

And with the current types, you'd get a warning for using it wrongly, whereas `Any | None` is bascially the same as having no types, and using proper coding style, you'd always have to check if the _conf is None or if the value in the _conf is None.

In general, returning Optional values and Any is considered bad practice, as the user has to guess what he'll get back and should (but hardly ever will) check his result.

puehringer: @dvmarkusvogl unfortunately, not all use cases are like this. Feel free to take a look at all repos if you need convincing. 

The point is that it returns a specific key (subsettings or **leaf**) in the JSON representation of the settings, which is 1:1 how the config worked before. I added this function only as compatibility layer, and compatibility layers should ideally have no breaking change, because that's kind of the purpose 🤗

So now, we can't change the behaviour so significantly, but it seems your issue can easily be solved by just passing a sensible default, i.e. {} if you expect a Dict. The entire point of the function is that I (as a tdp_core maintainer) don't know what you will get. 

So either pass the dict as your default, or even better, use the non-legacy pydantic model approach directly like settings.tdp_core.mongo.host for a typed, modern, recommended way. What is stopping you from using/updating to the latter approach?

puehringer:
@dvmarkusvogl This [commit](https://github.com/datavisyn/tdp_core/pull/673/commits/58bb997a29f68db2a9f8c6eb8fa09520264a4811) now adds tests (the contract so to say) and also how to properly unnest the models (for `tdp_core.security...` for example. That's what I would expect from all your settings. Please close this PR if all your questions/concerns are answered.

puehringer:
Closed as it is not an intended feature. Please reopen/create an issue if it causes trouble.

PR #707, opened 2022-05-02
Scatterplot Improvements

dvzacharycutler:


src/vis/bar/BarVis.tsx
```diff
@@ -100,6 +115,9 @@ export function BarVis({ config, optionsConfig, extensions, columns, setConfig,
         itemclick: false,
         itemdoubleclick: false,
       },
+      font: {
+        family: 'Roboto',
```
thinkh, line 119:
```suggestion
        family: 'Roboto, sans-serif',
```

I think we should change it to the same value that we have in the [_base.scss file](https://github.com/datavisyn/tdp_core/blob/f7c47fedd5dc41ab3d538b6656c42e3775eb3227/src/scss/base/_base.scss#L3). I'm not sure if it works here and if the syntax is correct. Please adapt for the other visualizations accordingly.

src/vis/sidebar/CloseButton.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export function CloseButton({ closeCallback }: { closeCallback: () => void }) {
+  return (
+    <div className="position-absolute end-0 top-0">
+      <button onClick={() => closeCallback()} className="btn btn-primary-outline" type="button">
+        <i className="fas fa-times" />
```
thinkh, line 7: Please replace with [Bootstrap close button](https://getbootstrap.com/docs/5.1/components/close-button/) that we use everywhere.

src/vis/scatter/utils.ts
```diff
@@ -208,11 +233,15 @@ export async function createScatterTraces(
...
-                      selected[v.id] ? '#E29609' : colorCol.type === EColumnTypes.NUMERICAL ? numericalColorScale(v.val as number) : scales.color(v.val),
-                    )
+                  ? hasSelected
+                    ? colorCol.resolvedValues.map((v) =>
+                        selected[v.id] ? (colorCol.type === EColumnTypes.NUMERICAL ? numericalColorScale(v.val as number) : scales.color(v.val)) : '#2e2e2e',
+                      )
+                    : colorCol.resolvedValues.map((v) =>
+                        colorCol.type === EColumnTypes.NUMERICAL ? numericalColorScale(v.val as number) : scales.color(v.val),
+                      )
                   : xCurr.resolvedValues.map((v) => (selected[v.id] ? '#E29609' : '#2e2e2e')),
```
thinkh, line 243: Should we extract the repeating color values to some config/constant file? Adding some variable names would also provide some documentation.

thinkh (changes requested):
Thanks for your changes. In general it looks already good to me. I have some minor suggestions for improvements that we could apply and discuss.

src/vis/bar/BarVis.tsx
dvzacharycutler: Done, seems to work everywhere

src/vis/sidebar/CloseButton.tsx
dvzacharycutler: done

src/vis/scatter/utils.ts
dvzacharycutler: Yes, agreed. Extracted them to a constants.ts file and changed it here as well as in the strip plots

thinkh (approved):
Thanks for your changes. Looks good to me.

PR #715, opened 2022-05-06
Remove Vis ranking listeners

oltionchampari:
Draft PR for now, will investigate further and test with multiple Vis Panels / rankings


puehringer:
@dvzacharycutler @oltionchampari Please finish this PR, as it causes performance issues if using the application for too long. In general, the problem is that the listeners are always added, however never removed, such that events are triggered multiple times (including the render).

puehringer:
@dvzacharycutler @oltionchampari See my comment above, I think this is still relevant?

oltionchampari:
@puehringer I am not sure what is there still to do here apart from some testing, this looks already good to me?

PR #718, opened 2022-05-12
LineUp: Default value for min/max in `numberCol()` and fix add column dialog in Ranking.tsx

dvchristianbors:
Fix two minor issues for Reprovisyn:
1. dialog text not showing
2. lineup tries to write read-only number column domain


thinkh (approved):
Tested in a call with @dvchristianbors. Looks good for me. 👍 

Amends PR #604.

PR #720, opened 2022-05-13
Adding proxy view react component

dvzacharycutler:
Added a very simple react component for proxy views. Used in reprovisyn for this pr https://github.com/datavisyn/reprovisyn/pull/383. 

The entire visyn view does not exist in tdp_core because there are different ways to convert from one mapping type to another across applications (or mostly just different in reprovisyn).


thinkh (approved):
Looks good to me.

PR #721, opened 2022-05-16
Refactor pushNotification to take a ReactNode

dvdanielamoitzi:
notification-handler


puehringer:
@oltionchampari Could you test this with tdp_comments? It should simply show the `Successfully saved|edited| ...` notifications, without any visible changes coming from this PR.

oltionchampari (changes requested):
Everything looks fine overall, one small issue only 
The ellipsis no longer shows in comments
image

see https://github.com/datavisyn/tdp_comments/blob/0dd645d6b25f0eee91a0488c627ba461aa916d65/src/ui/CommentUI.ts#L419
We should update the comments impl or fix this here before merging this.

puehringer:
@oltionchampari very good point indeed, to make it properly backwards compatible we would have to render the content as `innerHtml` via https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml in case it is a string. 
@dvdanielamoitzi could you take a look at it?

dvdanielamoitzi:
That's a good point Oli! I'll take a look at it :+1:

puehringer:
@oltionchampari @dvdanielamoitzi After a quick thought I would actually prefer to remove another attack vector of XSS and refactor the usages in the corresponding repositories. That shouldn't be much effort, right @oltionchampari?

thinkh:
> @oltionchampari @dvdanielamoitzi After a quick thought I would actually prefer to remove another attack vector of XSS and refactor the usages in the corresponding repositories. That shouldn't be much effort, right @oltionchampari?

@puehringer what is your suggested refactoring? Just replacing `&hellip;` with `...` is not the solution as there might be other characters that cannot be easily replaced.

oltionchampari:
> @oltionchampari @dvdanielamoitzi After a quick thought I would actually prefer to remove another attack vector of XSS and refactor the usages in the corresponding repositories. That shouldn't be much effort, right @oltionchampari?

So what do you propose? In my case I use react-html-parser to render formatted text that comes directly from elastic search the db

puehringer:
@thinkh @oltionchampari You can simply pass JSX instead: `<>Some text &hellip;</>`. React can handle those special characters, but it cannot handle them when it receives them as string.

thinkh:
> @thinkh @oltionchampari You can simply pass JSX instead: `<>Some text &hellip;</>`. React can handle those special characters, but it cannot handle them when it receives them as string.

Ok, I did not know this. However, our language strings are coming from I18N or from an API and they are not hard coded like in your example. Does this work as well?

oltionchampari:
> @thinkh @oltionchampari You can simply pass JSX instead: `<>Some text &hellip;</>`. React can handle those special characters, but it cannot handle them when it receives them as string.
That should not be a lot of effort.

puehringer:
> > @thinkh @oltionchampari You can simply pass JSX instead: `<>Some text &hellip;</>`. React can handle those special characters, but it cannot handle them when it receives them as string.
> 
> Ok, I did not know this. However, our language strings are coming from I18N or from an API and they are not hard coded like in your example. Does this work as well?

No, it won't work as it prevents XSS "by default". So any raw string is escaped and therefore shows _incorrectly_. I think the best solution here is to check whether the incoming value is a string, then use "the old way", and if it is a ReactNode then use the ReactDOM render way. Can you check this @dvdanielamoitzi? Ideally, we also add unit-tests testing exactly this functionality.

dvmoritzschoefl:
@puehringer @oltionchampari I think we can just ReactDOM.render + the DOMParser?
https://codesandbox.io/s/react-playground-forked-8ecny8?file=/index.js

puehringer:
I think this change is obsolete, we should be using https://mantine.dev/others/notifications/ instead.

PR #722, opened 2022-05-18
Proxy view simplifying

dvzacharycutler:
Removing some of the logic from the tdp_core proxy view, since its easier to have that logic in the visynview that uses this component instead.


thinkh (approved):
Looks good to me.

PR #737, opened 2022-06-21
Added d3v7 as dependency

puehringer:
Closes https://github.com/datavisyn/visyn2025/issues/134

Added `d3@7` as `d3v7` (incl. types) to the package.json. The `@types/d3v7` was added to the `dependencies` because then consumers of this library also get the types installed (as `devDependencies` are not installed transitively).
Also, removed all other `d3-*` dependencies and used `d3v3` alias instead. 

How to test in other applications: 

1. Switch tdp_core to this branch (in the workspace and/or package.json of the app)
2. Do a complete reinstall (yo phovea:update, npm i, ...)
3. Check if all the d3 imports still work
5. Optional: already upgrade to d3v7
   1. Remove all `d3` references in the package.json, i.e. to use the one coming with tdp_core
   2. Change import to `import * as d3v7 from 'd3v7';`
7. Create a branch and PR with this change.


puehringer:
d3v7 is already on tdp_core#develop now.

PR #749, opened 2022-06-30
Fix `ARankingView` typings

thinkh:
### Summary

The typings got more strict with PR https://github.com/datavisyn/tdp_core/pull/725, but the one for `rebuild` and `reloadData` were missing and caused TS errors in rankings of tdp_publicdb, which extend `ARankingView`.


src/lineup/ARankingView.ts
```diff
@@ -77,13 +77,13 @@ export abstract class ARankingView extends AView {
    * clears and rebuilds this lineup instance from scratch
    * @returns {Promise<any[]>} promise when done
    */
-  protected rebuild = debounceAsync(() => this.rebuildImpl(), 100);
+  protected rebuild: () => Promise<void> = debounceAsync(() => this.rebuildImpl(), 100);
```
puehringer, line 80: The depicted typings do not match with the `@returns` above, what is the correct return value then?

thinkh: Good point. I fixed it.

puehringer (approved):

PR #752, opened 2022-07-08
Fix incorrect route for server buildInfo.json

puehringer:
## Summary

When we migrated to FastAPI, we didn't append the `buildInfo.json` in the correct path (which should be `/api/buildInfo.json`). I updated this path and also added error handing if the route/backend might not be available.


thinkh (approved):
Looks good to me. I tested the route `api/buildInfo.json` and it works in my tests.

PR #753, opened 2022-07-13
Export initialize.ts for external Storybook instance

thinkh:
We need to run Storybook in Ordino and want to initialize the library as it is done for Storybook in tdp_core. Therefore, we need to export the `initializeLibrary()` function on package scope.


puehringer (commented):
This should be fine as the `initialize.ts` does not have any side-effects and only exports a single function. One question remains: It is now also importing `phovea.ts` (via `import reg from ...`). I don't think that was the case before, can you think of any problems that might arise from that?

thinkh:
I see your point. I think, in worst case it would import and register the content of the phovea.ts twice. But do you know if this happens already when a file is exported (= part of the bundle)? Or does this happen, when I actually import the function of the initialize.ts elsewhere?

puehringer:
Well, it will not register anything twice (as neither the initialize nor the phovea function are called anywhere), but it will import any files mentioned in the phovea.ts. That's the key difference I see, although that should be unproblematic.

PR #754, opened 2022-07-18
Set modal focus based on tour visiblity and replace Bootstrap jQuery modal with plain JS modal

thinkh:
Fixes Caleydo/tdp_bi_bioinfodb#1371

### Summary of changes

* Set modal `focus` flag based on tour's visibility
   * Set focus to `false` if a tour is visible to avoid focus conflicts between this modal and the tour dialog that automatically focus on the next button.
   * Otherwise the flag is set to `true` and keep the focus on this dialog.
* Replace Bootstrap jQuery modal with plain JS modal
* Fix `backdrop` property in dialog constructor (was commented out before)
* Replace deprecated `evt.keyCode === 27` with `evt.key === 'Escape'`


dvvanessastoiber (approved):
Works in my tests with the branch from [this PR](https://github.com/Caleydo/tdp_publicdb/pull/194).

puehringer (approved):

PR #787, opened 2022-09-21
feat: adding Mantine `VisynApp`

dvzacharycutler:
Closes https://github.com/datavisyn/reprovisyn/issues/560

### Developer Checklist (Definition of Done)

**Issue**

- [x] All acceptance criteria from the issue are met
- [x] Tested in latest Chrome/Firefox

**UI/UX/Vis**

- [ ] Requires UI/UX/Vis review
  - [ ] Reviewer(s) are notified (_tag assignees_)
  - [ ] Review has occurred (_link to notes_)
  - [ ] Feedback is included in this PR
  - [ ] Reviewer(s) approve of concept and design

**Code**

- [x] Branch is up-to-date with the branch to be merged with, i.e., develop
- [x] Code is cleaned up and formatted
- [ ] Unit tests are written (frontend/backend if applicable)
- [ ] Integration tests are written (if applicable)

**PR**

- [x] Descriptive title for this pull request is provided (will be used for release notes later)
- [x] Reviewer and assignees are defined
- [ ] Add type label (e.g., *bug*, *feature*) to this pull request
- [ ] Add release label (e.g., `release: minor`) to this PR following [semver](https://semver.org/)
- [x] The PR is connected to the corresponding issue (via `Closes #...`)
- [x] [Summary of changes](#summary-of-changes) is written

### Summary of changes

- Adding Mantine VisynApp component that was first used in Sightline. Adding fontawesome imports to dependencies, and mantine imports to devDependencies

### Additional notes for the reviewer(s)

---
Thanks for creating this pull request 🤗


src/visynApp/VisynApp.tsx
```diff
@@ -0,0 +1,56 @@
...
+  loginMenu?: JSX.Element;
+  children?: React.ReactChild;
+  headerHeight?: number;
+  appName: string;
+}) {
+  const { status } = useInitVisynApp();
+
+  return status === 'success' ? (
+    <AppShell
+      styles={{ root: { height: '100%' }, body: { height: `calc(100% - ${headerHeight}px)` }, main: { minHeight: '0px' } }}
```
thinkh, line 45: I would leave the definition of the styles to app, because you don't need this for example in [Sightline](https://github.com/datavisyn/sightline/blob/d7f67b1aebd92fe40d135209452fb04e9a926526/src/app/visynApp/VisynApp.tsx#L41).

```jsx
<VisynApp styles={{ root: { height: '100%' }, body: { height: `calc(100% - ${headerHeight}px)` }, main: { minHeight: '0px' } }}></VisynApp>
```

Otherwise: Why is the headerHeight configurable, when there is a constant in VisynHeader.tsx?

```ts
const HEADER_HEIGHT = 50;
```

Then the variable `headerHeight` can be removed and set to 50, right?

src/visynApp/login/LoginDialog.tsx
```diff
@@ -0,0 +1,76 @@
...
+  hasError?: boolean;
+  setError: (s: string) => void;
+  appName: string;
+}) {
+  return (
+    <Modal withCloseButton={false} opened={show} onClose={() => null} title={null}>
+      <Container fluid>
+        <Stack>
+          <Center>
+            <Title order={4}>{`Welcome to ${appName}`}</Title>
```
thinkh, line 40: Please use I18N

src/visynApp/login/LoginDialog.tsx
```diff
@@ -0,0 +1,76 @@
...
+          </Center>
+          <Divider />
+          <Center>
+            <Text mb="lg">Please log in to continue</Text>
```
thinkh, line 44: Please use I18N

src/visynApp/login/VisynLoginForm.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <ActionIcon onClick={() => setIsShowPassword(!isShowPassword)}>
+              {isShowPassword ? <FontAwesomeIcon icon={faEye} /> : <FontAwesomeIcon icon={faEyeSlash} />}
+            </ActionIcon>
+          }
+          required
+        />
+      </Stack>
+      <Group position="right">
+        <Button fullWidth={false} mt="md" type="submit" className="btn btn-primary">
+          Log In
```
thinkh, line 44: Please use I18N

thinkh (changes requested):
Thanks for adding the VisynApp and related components here. Please see my comments below.

One additional remark:

I suggest to move the datavisyn_white.svg to the assets folder. (GitHub does not allow to add comments to the SVG preview).

package.json
```diff
@@ -54,6 +54,10 @@
   },
   "dependencies": {
     "@fortawesome/fontawesome-free": "^6.1.1",
+    "@fortawesome/fontawesome-svg-core": "^6.1.2",
+    "@fortawesome/free-regular-svg-icons": "^6.1.2",
+    "@fortawesome/free-solid-svg-icons": "^6.1.2",
+    "@fortawesome/react-fontawesome": "^0.2.0",
```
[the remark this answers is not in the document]
puehringer: If `@fortawesome/react-fontawesome` does not install a different version of fontawesome (or bundle it twice for some reason), I don't mind these helper functions (I would actually use them I guess).

package.json
```diff
@@ -95,6 +99,10 @@
   },
   "devDependencies": {
     "@babel/core": "^7.17.7",
+    "@emotion/react": "^11.10.0",
+    "@mantine/core": "5.3.0",
+    "@mantine/form": "^5.1.1",
+    "@mantine/hooks": "^5.0.2",
```
puehringer, line 105: Shouldn't we pin all of them to `~5.3.0` then? Mixing it doesn't make too much sense I guess. And why not include patch versions?

thinkh: @dvzacharycutler I agreed with Michael that we use `~5.3.0` for all Mantine packages. Please change.

package.json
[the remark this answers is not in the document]
thinkh: @dvzacharycutler Please make a production build with `yarn run build:tdp_core` and check if the bundle contains all svg icons or just the used subset, i.e. your case only one or two icons for this PR. I would expect that the bundle size shouldn't a lot bigger than without the new packages.

src/visynApp/VisynApp.tsx
dvzacharycutler: The idea is by default we use the VisynHeader, but since users can override the header with any component they need, they should also be able to pass a custom header height. Now the next question, "if its a custom header, why do you need its height?". Great question! this is something weird about mantine (or maybe I just don't understand), but if you don't do this calculation for the height that I do on the body styles your components with normal 100% height will overflow. This is actually also why I included it in here, since I believe all applications will need something similar, and it might be hard to figure out why, as it took me a while in sightline.

src/visynApp/login/VisynLoginForm.tsx
dvzacharycutler: done

src/visynApp/login/LoginDialog.tsx
dvzacharycutler: done

src/visynApp/login/LoginDialog.tsx
dvzacharycutler: done

package.json
dvzacharycutler: Done

package.json
```diff
@@ -95,6 +99,10 @@
   },
   "devDependencies": {
     "@babel/core": "^7.17.7",
+    "@emotion/react": "^11.10.0",
+    "@mantine/core": "5.3.0",
```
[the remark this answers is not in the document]
dvzacharycutler: Resolved below

PR #820, opened 2023-01-19
Enable Custom Vis in LineUp

dvmoritzschoefl:


src/lineup/Ranking.tsx
```diff
@@ -152,6 +154,7 @@ const defaults: IRankingOptions = {
 export interface IRankingOptions extends IARankingViewOptions {
   mode: EViewMode;
   enableCustomVis: boolean;
+  idField: string;
```
puehringer, line 157: Could we make this optional to ensure backwards compatibility?

puehringer (approved):

PR #825, opened 2023-02-06
feat(vis): allow column domains to be set

dvzacharycutler:
Closes https://github.com/datavisyn/aelixir/issues/325

### Developer Checklist (Definition of Done)

**Issue**

- [ ] All acceptance criteria from the issue are met
- [ ] Tested in latest Chrome/Firefox

**UI/UX/Vis**

- [ ] Requires UI/UX/Vis review
  - [ ] Reviewer(s) are notified (_tag assignees_)
  - [ ] Review has occurred (_link to notes_)
  - [ ] Feedback is included in this PR
  - [ ] Reviewer(s) approve of concept and design

**Code**

- [ ] Branch is up-to-date with the branch to be merged with, i.e., develop
- [ ] Code is cleaned up and formatted
- [ ] Unit tests are written (frontend/backend if applicable)
- [ ] Integration tests are written (if applicable)

**PR**

- [ ] Descriptive title for this pull request is provided (will be used for release notes later)
- [ ] Reviewer and assignees are defined
- [ ] Add type label (e.g., *bug*, *feature*) to this pull request
- [ ] Add release label (e.g., `release: minor`) to this PR following [semver](https://semver.org/)
- [ ] The PR is connected to the corresponding issue (via `Closes #...`)
- [ ] [Summary of changes](#summary-of-changes) is written

### Summary of changes

- Added ability to assign the initial domain for numerical columns, affecting scatterplots only

### Additional notes for the reviewer(s)

-  
Thanks for creating this pull request 🤗


src/vis/interfaces.ts
```diff
@@ -169,6 +170,8 @@ export type PlotlyData = {
   xTickLabels?: string[];
   yTicks?: string[];
   yTickLabels?: string[];
+  xDomain?: [number, number];
```
puehringer, line 173: @dvzacharycutler the domain should support passing [0, undefined]. Then the typing should be [number | undefined, number | undefined].

dvzacharycutler: This is a bit tricky, since if we set these domains we have to set both numbers in the plotly layout. I can get [0, undefined] specifically to work, by using the clamp to 0 setting, but [1, undefined] wouldn't be possible without manually calculating maximum/minimum myself in Vis. That's probably not a huge deal though? Shouldn't happen frequently

puehringer: I guess it's a good thing to calculate min (if undefined) and/or max (if undefined). Should just be a few lines, and it won't be slower as plotly has to do it otherwise/it doesn't happen that often.

src/vis/scatter/utils.ts
```diff
@@ -26,6 +26,18 @@ export function isScatter(s: IVisConfig): s is IScatterConfig {
   return s.type === ESupportedPlotlyVis.SCATTER;
 }
 
+function calculateDomain(domain: [number | undefined, number | undefined], vals: number[]): [number, number] {
+  if (domain[0] && domain[1]) {
```
puehringer, line 30: @dvzacharycutler You will have to be more specific here, as 0 is a falsy value. I.e. you will need to explicitly check for `domain[0] == null`. 

Generally, I would rewrite this as follows: 

```js
return [domain[0] == null ? Math.min(...(vals as number[])) : domain[0], domain[1] == null ? Math.max(...(vals as number[])) : domain[1]]
```
If you need the `max + max / 20`, just extract the two as variables and return as array accordingly.

src/vis/scatter/utils.ts
```diff
@@ -26,6 +26,18 @@ export function isScatter(s: IVisConfig): s is IScatterConfig {
   return s.type === ESupportedPlotlyVis.SCATTER;
 }
 
+function calculateDomain(domain: [number | undefined, number | undefined], vals: number[]): [number, number] {
+  if (!domain || (domain[0] !== undefined && domain[1] !== undefined)) {
```
puehringer, line 30: @dvzacharycutler  This check can't be correct? I.e. if domain is null, you would still return `[domain[0], domain[1]]`?

puehringer (approved):

PR #835, opened 2023-02-15
Migration to visyn_core

puehringer:
This is the migration to [visyn_core](https://github.com/datavisyn/visyn_core), the new "light" core without bootstrap, jquery, fontawesome scss (use [this](https://github.com/datavisyn/visyn_core/blob/d69ac110780d55e753d1d1131abb521454b7a73c/src/app/header/VisynHeader.tsx#L2-L5) instead), ...  
New(er) and actively developed apps should not at all require `tdp_core`, as this is meant for legacy applications. 

### Migration guide
* For any repo, first check if there is a `visyn_core` branch already, if not, create one
* If your repo/application might work without tdp_core (i.e. doesn't use any of the "old" features like clue, bootstrap, ...), then we continue with visyn_core only (adding tdp_core back is easy). 
* Remove tdp_core from your package.json and requirements.txt, and add visyn_core
  * package.json: `"visyn_core": "git+ssh://git@github.com:datavisyn/visyn_core#develop"` and `"visyn_scripts": "git+ssh://git@github.com:datavisyn/visyn_scripts#visyn_core"`
  * requirements.txt: `visyn_core@git+https://github.com/datavisyn/visyn_core.git@develop#egg=visyn_core`
*  Do a fresh install
   * `rm -rf .venv && python -m venv .venv && source .venv/bin/activate && make develop`
   * `rm -rf yarn.lock node_modules && yarn`
* Change imports and check if everything works
  * I.e. for the backed, search globally for `tdp_core` for `*.py` files, and replace that with `visyn_core`. Then run `make all`. If you have unit tests/pyright, you should see where things fail. If everything passes, it works already! 
    * Some imports are not part of visyn_core anymore, like `no_cache` (not required anymore, remove it), namedsets, graphs. If you really need them, you need to add tdp_core back and change the missing imports to tdp_core. 
  * I.e. for the frontend, search globally for `'tdp_core'` for `*.ts,*.tsx` files, and replace that with `'visyn_core'`. Run `yarn run compile:watch` and you should probably see some TS errors already. If an import is also not found in visyn_core (like `Dialog, Errors, PHOVEA_*, ...`), you need to either refactor it (if it only occurs once), or add tdp_core (`"tdp_core": "git+ssh://git@github.com:datavisyn/tdp_core#visyn_core"`) back to your package.json (and hence remove the `visyn_core` entry as it comes with `tdp_core`)  and change the import back to tdp_core. Also, don't forget to add back the import for tdp_core in the phovea.ts. 
    * Important: In your phovea_registry.ts, you should 1:1 reflect your package.json. I.e. if you have both tdp_core and visyn_core, you need to have `import 'visyn_core/phovea_registry';` and `import 'tdp_core/dist/phovea_registry';` (notice the missing /dist/ in visyn_core). 
* If your frontend compiles and your backend works, continue by starting the application via `yarn start` and `make start` (and any db if needed). 
  * If you now get import errors by webpack (not typescript), resolve them. If you don't know how, let us know immediately as it could hinder the migration. If you get weird errors regarding ESM imports, also let us know. 
* Test your application. If you encounter issues, let us know here/in Slack.
* If everything works as expected, create a PR titled `Migration to visyn_core` (with the `visyn_core` branch), add @puehringer as reviewer, and link this ticket. Thanks! 

Example migration with both visyn_core and tdp_core: https://github.com/datavisyn/tdp_comments/pull/278
Example migration with visyn_core only :tada: : https://github.com/datavisyn/tdp_configuration_manager/pull/79
Example migration of an app: https://github.com/datavisyn/dv_cime/pull/93
Example migration of the app_template: 
https://github.com/datavisyn/app_template/commit/94eb7f7d2d3d0efdea6da34e2180a01c574dfeaa

## Advanced topics
If your application is able to run with `visyn_core` only, you can go even further by using deep imports to simplify tree-shaking. Every package is exported explicitly from `visyn_core`, i.e. if you need the `Vis`, you can import from either `visyn_core` or `visyn_core/vis`. The second is much better, as it avoids importing the entirety of `visyn_core`, thus drastically reducing bundle size. However, it is important to notice that your entire application must avoid `visyn_core` imports then for tree-shaking to work. Therefore, search for `'visyn_core'` imports, and replace them with the corresponding package like `'visyn_core/vis'`, `'visyn_core/app'`, ...


src/views/ViewUtils.ts
```diff
@@ -184,23 +174,6 @@ export class ViewUtils {
     });
   }
 
-  static findVisynViews(idType?: IDType): Promise<VisynViewPluginDesc[]> {
```
oltionchampari, line 187: @puehringer  How do we achieve the same thing again, since this got removed, and i did not see anything similar in visyn_core.
`PluginRegistry.listPlugins()` and then filter?

puehringer: Yep, simply move this function to your codebase (or use PluginRegistry.listPlugins + filter).

